PR #1, opened 2018-03-07
[PPTECHDOC-19 PPTECHDOC-7] Added structure and init files, added SDK …

JacekClearcode:
…documentation


_static/css/custom.css
```diff
@@ -0,0 +1,15 @@
+.wy-side-nav-search {
+	background-color: #dd1144;
+}
+
+dl.post, dl.patch, dl.delete, dl.get, dl.put {
+	display: none
+}
+
+[sticky-sidebar][_ngcontent-vyl-1], .menu-content {
+	max-height: calc(100% - 200px) !important;
```
RomanSek: Do we have any limitations regarding supported browsers? This is quite new syntax.

_static/js/custom.js
```diff
@@ -0,0 +1,17 @@
+(function () {
+    window.onload = function () {
+        var isApiPresent = document.getElementsByTagName('redoc').length;
+        if(isApiPresent) {
+            window.addEventListener("hashchange", function(){
+                var locationHash = location.hash.split('#')[1],
```
RomanSek: I'm a bit worried what will happen if `location.hash` will contain extra `#` character, but I'm not sure if it's possible to generate such link using `redoc`.

_static/js/custom.js
```diff
@@ -0,0 +1,17 @@
...
+                    elementToScroll = null;
+                if(locationHash.split('/')[0] !== 'section'&& locationHash.split('/')[0] !== 'tag' ) {
```
RomanSek: Assign this value to a variable so you don't have to generate it twice.

_static/js/custom.js
```diff
@@ -0,0 +1,17 @@
...
+                    elementToScroll = document.querySelector('[href="'+ location.hash +'"]');
+                }
+                if(elementToScroll){
```
RomanSek: I don't see a point to this `if` statement. Can't you scroll as soon as `elementToScroll` is set (in previous `if` statement)?

conf.py
```diff
@@ -0,0 +1,217 @@
...
+# You can specify multiple suffix as a list of string:
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'PPMS-PDFD'
```
RomanSek: Project name is displayed in documentation. Is this official name?

index.rst
```diff
@@ -0,0 +1,15 @@
...
+   sphinx-quickstart on Tue Dec 12 14:22:39 2017.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Piwik PRO Developers documentation!
+=====================================
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Mobile SDK documentation
```
RomanSek: I was trying to avoid setting captions for each section. I thought that maybe we could use them to separate apps, but only if it would be necessary. I would start without captions.

index.rst
```diff
@@ -0,0 +1,15 @@
+.. PPMS-PDFD documentation master file, created by
+   sphinx-quickstart on Tue Dec 12 14:22:39 2017.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Piwik PRO Developers documentation!
+=====================================
+
+.. toctree::
+   :maxdepth: 2
```
RomanSek: We should be careful with this setting to not overload menu. I would start with it set to 1.

index.rst
```diff
@@ -0,0 +1,15 @@
...
+   :caption: Mobile SDK documentation
+   :name: test
+   :glob:
+
+   sdk/*
```
RomanSek: I suggested to not use global linking. I would like to specify whole structure explicite to avoid situations where files are in unexpected sequence. This is also my strategy to hide deeper pages using `maxdepth` option.
Additionally `sdk` contains `md` files, but `config.py` expects `rst`. Does it work?

_static/js/custom.js
JacekClearcode: It's not possible.

_static/css/custom.css
JacekClearcode: Added from CSS that are used in docs

_static/js/custom.js
JacekClearcode: This is more like POC, it will be fixed in another ticket

_static/js/custom.js
JacekClearcode: This is more like POC, it will be fixed in another ticket

conf.py
JacekClearcode: No.

index.rst
JacekClearcode: I will leave this caption here as this is complete documentation for SDK section

index.rst
JacekClearcode: I will leave it for now, will check if this will overload menu as we will enter more documentations

index.rst
JacekClearcode: This was suggested by RTD tutorial.

PR #2, opened 2018-03-09
[PPTECHDOC-20] Added files and fixed indexes

JacekClearcode:


audience_manager/index.rst
```diff
@@ -0,0 +1,14 @@
+.. PPMS-PDFD documentation master file, created by
```
RomanSek: You don't need this comment in subdir indexes.

audience_manager/index.rst
```diff
@@ -0,0 +1,14 @@
...
+   sphinx-quickstart on Tue Dec 12 14:22:39 2017.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Piwik PRO Audience Manager documentation!
```
RomanSek: This is title of this section of documentation. In this case it's Audience manager app. You can also add short description of the app before `toctree`.

audience_manager/index.rst
```diff
@@ -0,0 +1,14 @@
...
+====================================================
+
+.. toctree::
+  :maxdepth: 2
+  :glob:
```
RomanSek: You don't need `glob` if you don't use wildcards.

sdk/index.rst
```diff
@@ -0,0 +1,14 @@
...
+   sphinx-quickstart on Tue Dec 12 14:22:39 2017.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Piwik PRO SDK documentation!
+=======================================
+
+.. toctree::
+   :maxdepth: 2
+   :name: test
```
RomanSek: Im still not sure how this `name` is used. I think you can skip it unless you need it.

audience_manager/index.rst
```diff
@@ -0,0 +1,14 @@
+.. PPMS-PDFD documentation master file, created by
+   sphinx-quickstart on Tue Dec 12 14:22:39 2017.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Piwik PRO Audience Manager documentation!
+====================================================
+
+.. toctree::
+  :maxdepth: 2
```
RomanSek: This value will control how deep will be ToC and displayed sidebar links. You should check how they look after generation and adjust it accordingly.

index.rst
```diff
@@ -7,9 +7,13 @@ Welcome to Piwik PRO Developers documentation!
...
-   :glob:
+   :maxdepth: 1
 
-   sdk/*
\ No newline at end of file
+   sdk/index
+
+
+
+.. toctree::
```
RomanSek: You don't need separate `toctree` if you don't use captions and names. Just link all documents in one ToC.

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
+.. highlight:: js
+.. default-domain:: js
+
+Audience Manager JavaScript API
+===============================
+This API provides JavaScript access to multiple resources.
```
zaan: What actually means "access to multiple resources"? It's not clear to me. The API, in general, allows to push and read profile related data and we should communicate that.

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Loading snippet
+---------------
+Add following snippet on a page to start using Audience Manager JS API.
```
zaan: "Following snippet" suggests that below the paragraph the snippet is located while it's 3 sections below. It was confusing to me when I was reading it for the first time.

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+
+
+Commands
+--------
+Following commands work in context of current browser user profile. Additionally these commands require communication
+with server so they are asynchronous. Callback functions are used to provide response or error information.
+
+List audiences
+``````````````
+Gets list of user audiences.
```
zaan: I would be careful with using "user" as it's not 100% clear. Maybe we should be using either "visitor" or "user profile".

The function returns a list of audiences the visitor belongs to (not _his_ audiences as specified above).

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+
+.. function:: onRejected(error)
+
+    Function executed on error.
+
+    :param string error: Error code. Example::
+
+        'server_error'
+
+Check audience
```
zaan: It's general and does not say what the function actually does. I suggest following header:
Checking the visitor membership in the audience

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+
+        'server_error'
+
+Get attributes
+``````````````
+Gets user attributes.
+
+.. warning::
+
+    Only white-listed attributes are available this way.
```
zaan: Where can I manage the whitelist? Why are only whitelisted attributes available? Take a look at backend API docs for an inspiration: http://am.devtools.piwik.pro/#tag/attributes%2Fpaths%2F~1audiences~1public~1api~1profiles~1%7Banalytics_visitor_id%7D~1attributes~1%2Fget

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+
+    Function executed on error.
+
+    :param string error: Error code. Example::
+
+        'server_error'
+
+Update attributes
+`````````````````
+Updates user attributes.
```
zaan: Questions that come to my mind when I read this short sentence:
- Do I update only values of existing attributes?
- Can I create a new attribute?
- If I will use attribute name from "Analytics" section (i.e. location), will that be overwritten?

Those descriptions are very short and should give some background so that developers understand AM behavior better. See backend API docs for inspiration: http://am.devtools.piwik.pro/#tag/attributes%2Fpaths%2F~1audiences~1tracker~1attributes~1%2Fpost

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+    Object containing attributes to update. Every key and value should be a string. Example::
+
+        {
+            "favourite_color": "black",
+            "drink: "Martini"
+        }
+
+.. data:: options
+
+    **Optional** object that lets execute command with extra parameters. Example::
```
zaan: If I will use any identifiers in "options" and it's not yet set in the profile, will it be saved in the profile? For example, I will set an email option but the visitor's profile doesn't have the email set so far - is it going to be saved in the profile? It would be good to describe the behavior, what happens (or doesn't happen) with those options.

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+        value should be unique user identifier provided by the App. It may be user handle or internal user ID. Example::
+
+            "jbond"
+
+    .. attribute:: device_id
+
+        If application has access to device id this option will let Audience Managed better identify user across
+        sessions. Example::
+
+            "007"
```
zaan: In my opinion, the example should be similar to real device ID. For sure none of the device IDs consists of 3 characters.

audience_manager/api.rst
```diff
@@ -0,0 +1,213 @@
...
+
+    .. attribute:: email
+
+        If application can track user by his email this option will let Audience Managed better identify user across
```
zaan: Typo: "Audience Manage**d**" instead of "Audience Manager"

audience_manager/form-tracker.rst
```diff
@@ -0,0 +1,101 @@
+.. highlight:: js
+.. default-domain:: js
+
+Audience Manager Form Tracker API
+=================================
+Form Tracker allows to gather data from all forms on your site in one place.
```
zaan: It's unclear and confusing. It should say directly that it captures data from forms and saves it as a profile attribute of the visitor who submitted the form. The above sentence says that form data is saved and I can interpret it in very different ways.
Documentation should also explain how attributes are created (from names of fields), inform where attributes are saved and in what cases it can overwrite other existing attributes (i.e. those created based on custom variable or dimension). 
I think it would be also a good idea to link help center article that explains form tracking behavior.

audience_manager/form-tracker.rst
```diff
@@ -0,0 +1,101 @@
...
+Form tracker is trying to automatically detect fields containing private information and exclude them. Following fields
+are always removed from tracked data:
+
+* "password" or "hidden" inputs
+* Credit card number
+* Credit card validation code
+
+Loading snippet
+---------------
+Add following snippet on a page to start using Audience Manager Form Tracker.
```
zaan: The same issue as for JS API docs - it says "following snippet" and I don't see that snippet as it's much lower and I need to scroll down to find it.

Template tag in tag manager is not mentioned here at all while it is a preferred method of injecting form tracking code, the snippet is an alternative.

PR #7, opened 2018-03-26
Feature/pptechdoc 24 piwik js tracking api

JacekClearcode:


analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,842 @@
...
+
+    _paq.push(["setCustomVariable", index, name, value, scope]);
+
+.. data:: index
+
+    Number from 1 to 5 where variable is stored
+
+.. data:: name
+
+    Name of the variable
```
RomanSek: I'm confused. I thought that variable is stored in slot (configured in Piwik before first use), but here we're sending index and name of custom variable. What happens when name doesn't match configured name of variable? Is it overwritten, ignored or does each name have separate key-value inside each index? I don't see it used in other methods for custom variables (delete and read). What's the point of this value?

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,842 @@
...
+
+.. data:: value
+
+    Value of the variable
+
+.. data:: scope
+
+    Scope of the variable, "visit" or "page"
```
RomanSek: We need to describe how those scope options work. I'm guessing that page variables are not stored anywhere and they are lost on page load, while visit is stored in cookie and used for duration of a session.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,842 @@
...
+
+
+.. warning::
+
+    Custom Variable with scope set on "visit" will be saved for visit, you don"t need to save it every page.
```
RomanSek: Typo: Quote error.

analytics/tracker_object.rst
```diff
@@ -0,0 +1,543 @@
+.. highlight:: js
+.. default-domain:: js
+
+Tracker Object Functions
+========================
+
+This document describes all functions available for Tracker object and how to create its instance.
```
glothriel: Here Tracker is uppercase and 6 lines below its lowercase, please decide one standard.

analytics/tracker_object.rst
```diff
@@ -0,0 +1,543 @@
...
+.. function:: trackContentInteractionNode(domNode, contentInteraction);
+
+    Function that tracks interaction within domNode. Can be used as a function inside onClick attribute.
+
+    :param domNode domNode: **Required** Any node in content block or the block itself - it won't be tracked if no content block will be found
+    :param string contentInteraction: **Optional** String containing name of interaction it can be anything ("click" etc). "Unknown" used as default.
+
+.. function:: trackContentInteraction(contentInteraction, contentName, contentPiece, contentTarget);
+
+    Function that content interaction using given data.
```
glothriel: Probably missing `track`

analytics/tracker_object.rst
```diff
@@ -0,0 +1,543 @@
...
+    :param string keyword: **Required** String containing keyword that was searched.
+    :param string/boolean category: **Required** String with category selected in search engine, can set it to false when not used.
+    :param number/boolean searchCount:  **Required** Number of results on the results page, can be set to false when not used.
+
+.. function:: enableHeartBeatTimer(delay);
+
+    When user will enter single page on visit we will assume that total time spent on website was 0 ms.
+Function will measure that time more accurately.
+
+    :param number delay: **Required** Time in seconds, when send another request with heartbeat, default ``30``
```
glothriel: This sentence doesn't sound correctly in English.

analytics/tracker_object.rst
```diff
@@ -0,0 +1,543 @@
...
+    Function that will track internal site searches.
+
+    :param string keyword: **Required** String containing keyword that was searched.
+    :param string/boolean category: **Required** String with category selected in search engine, can set it to false when not used.
+    :param number/boolean searchCount:  **Required** Number of results on the results page, can be set to false when not used.
+
+.. function:: enableHeartBeatTimer(delay);
+
+    When user will enter single page on visit we will assume that total time spent on website was 0 ms.
+Function will measure that time more accurately.
```
glothriel: Probably missing `this` before function. It's somewhat inconsistent, it does not measure, it enables this setting.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+    _paq.push(["trackLink", linkAddress, "link", dimensions]);
+
+.. data:: linkAddress
+
+    **Required** Address that link points to.
+
+.. data:: dimensions
+
+    **Optional**  :ref:`Custom dimension<Custom Dimensions>` that should be tracked with this action. Can be multiple dimensions.
+    Written as object property using ``dimension{ID}`` notation.
```
RomanSek: There's missing explanation what should be replaced in property name. It would be good to provide also few examples right away (unless example bellow is enough).

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+
+    **Required** String with css class that should be tracked instead of ``piwik_link``
+
+Example of usage::
+
+    _paq(["setLinkClasses", "track-this-link"]);
+
+If you want to use JS instead you can add ``trackLink`` function to element ``onClick`` attribute::
+
+    _paq.push(["trackLink", linkAddress, "link", dimensions]);
```
RomanSek: Why is "link" not a parameter? Does it always have to be this value? 

Update: I see that it's used with different value later on. I would declare it once and link to it when necessary. Another question is: This section describes outlinks, but this function is tracking "link". Is there no distinction between them? In that case section is tracking links and it so happens that by default it tracks only links to external domains.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+| exe   | flv | gif | gz  | gzip | hqx | jar | jpg | jpeg | js  | mp2  | mp3 | mp4     | mpg | mov |
++-------+-----+-----+-----+------+-----+-----+-----+------+-----+------+-----+---------+-----+-----+
+| movie | msi | msp | odb | odf  | odg | odp | ods | odt  | ogg | ogv  | pdf | phps    | png | ppt |
++-------+-----+-----+-----+------+-----+-----+-----+------+-----+------+-----+---------+-----+-----+
+| qt    | qtm | ra  | ram | rar  | rpm | sea | sit | tar  | tbz | tbz2 | tgz | torrent | txt | wav |
++-------+-----+-----+-----+------+-----+-----+-----+------+-----+------+-----+---------+-----+-----+
+| wma   | wmv | wpd | xls | xml  | z   | zip |     |      |     |      |     |         |     |     |
++-------+-----+-----+-----+------+-----+-----+-----+------+-----+------+-----+---------+-----+-----+
+
+You can add extension to this list using ``addDownloadExtensions`` function::
```
RomanSek: Is it a subsection?

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+.. data:: extensions
+
+    **Required** String containing extensions separated by ``|`` for example ``"mhj|docx"``
+
+
+Example of usage::
+
+    _paq.push(["addDownloadExtensions", "mhj|docx"]);
+
+This list can be rewrote using ``setDownloadExtensions`` function::
```
RomanSek: Is it a subsection?

Also this sentence is hard to read. Maybe something like this: "This list can be overwritten ...".

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+.. code-block:: html
+
+    <button onClick = function(){_paq.push(["trackLink", "http://www.example.com/example", "link"]);}>
+        Click me!
+    </button>
+
+Tracking Downloads
+``````````````````
+
+Following extensions are tracked as download by default:
```
RomanSek: There's missing explanation what is download tracking doing. From what I gather - it tracks clicks on links to files ending with specific extensions. I'm still not sure if file URL has to end on this extension or if it can also contain GET params (e.g. "http://example.com/file.zip?cachebuster=9831274"). 
I suspect that downloads gong through not matching redirects or HTML pages with URL ending with "zip" will fool this tracker. It would be good to give at lest cursory explanation how it works.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+.. data:: extensions
+
+    **Required** String containing extensions separated by ``|`` for example ``"7z|apk|mp4"``
+
+
+Example of usage::
+
+    _paq.push(["setDownloadExtensions", "7z|apk|mp4"]);
+
+To track clicking a link as an download using css class simply add ``piwik_download`` class to link element.
```
RomanSek: Is it a subsection?

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+.. data:: className
+
+    **Required** String with css class that should be tracked instead of ``piwik_download``
+
+Example of usage::
+
+    _paq(["setDownloadClasses", "track-this-link-for-download"]);
+
+
+If you want to use JS instead you can add ``trackLink`` function to element ``onClick`` attribute::
```
RomanSek: This function is declared a second time. Maybe it would be better to define it in one place and link to it as alternative to automatic link/download tracking.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+Example of usage
+
+.. code-block:: html
+
+    <button onClick = function(){_paq.push(["trackLink", "http://www.example.com/example.xrt", "download"]);}>
+        Click me!
+    </button>
+
+
+After each outbound link there is small time frame after which the file will download that will
```
RomanSek: I don't get it. Does it mean there is some kind of sync delay to ensure tracking before redirect? How does it work?

It's probably another sub-section.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+That time frame is set to 500ms by default. To modify it you can use ``setLinkTrackingTimer`` function::
+
+    _paq.push(["setLinkTrackingTimer" time]);
+
+.. data:: time
+
+    **Required** Time in ms between user interaction and downloading file.
+
+
+Disabling tracking
```
RomanSek: I would like to have it near CSS enable section, but that is separated between link and download sections. Can they be moved to one place (outside link and downloads sections)?

It looks like this function is reverse to 2 other specialized functions for links and downloads. I'm not sure what to do with it. In the future it may be better to have one function and one reverse function to keep it clean.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+^^^^^^^^^^^^^^^^^^
+User ID enables merging user data that is collected between many devices and browsers.
+
+You must provide unique user-id for every user. To set user ID for tracked data use ``setUserId`` function::
+
+    _paq.push(["setUserId", userID]);
+
+.. data:: userID
+
+    **Required** Unique, non empty string preserved for each user.
```
RomanSek: I  suspect "preserved" meant "Unique, non empty permanent ID of the user in application." It may be easier to use later these terms from glossary to uphold standard naming convention.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+
+Tracking domains and subdomains
```
RomanSek: This section looks like good candidate to move to separate document describing in full manual snippet installation since those configs will be a lot less popular.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Tracking single domain
+``````````````````````
+To track single domain name without tracking subdomains (or single subdomain) use default snippet code::
+
+    _paq.push(["setSiteId", 1]);
```
RomanSek: To be clear I would provide full snippet examples (HTML with `<script>` tag).

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+
+::
+
+    _paq.push(["enableCrossDomainLinking"]);
+
+Tracking subdirectories of domain in separate websites.
+```````````````````````````````````````````````````````
+To differentiate parts of website as another site for tracker user must do::
+
+    _paq.push(["setSiteId", Q]);
```
RomanSek: What's `Q`? It wasn't used in previous examples.

Update it's explained bellow, but it's confusing to read in this order.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+
+    _paq.push(["trackContentInteractionNode", domNode, contentInteraction]);
+
+.. data:: domNode
+
+    **Required** ny node in content block or the block itself - it won't be tracked if no content block will be found
+
+.. data:: contentInteraction
+
+    **Required** String containing name of interaction it can be anything ("click" etc). "Unknown" used as default.
```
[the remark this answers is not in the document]
JacekClearcode: Did you read the description? Im afraid you did not.  * just add this function as an eventListener for action you want*

And you can write anything you want in contentInteraction.

analytics/js_tracking_api.rst
JacekClearcode: where Q is unique site id. It was written BELOW

analytics/js_tracking_api.rst
JacekClearcode: I think this should still be in the same document for now.

analytics/js_tracking_api.rst
JacekClearcode: These things are written like that in the tracker object reference

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+
+    _paq.push([ function() {
+        var customVariable = this.getCustomVariable(index, scope );
+    }]);
+
+.. function:: getCustomVariable(index, scope)
+
+    :param number index: **Required** Number from 1 to 5 where variable is stored
+
+    :param string scope: **Required** Scope of the variable, "visit" or "page"
```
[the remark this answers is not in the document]
JacekClearcode: I would stay on a single document

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,859 @@
...
+```````````````````````````
+You can access custom dimension by providing function that will use ``getCustomDimension`` function::
+
+    _paq.push([ function() {
+        var customDimension = this.getCustomDimension(index);
+    }]);
+
+.. function:: getCustomDimension(index, scope)
+
+    :param number index: **Required** Index of custom dimension
```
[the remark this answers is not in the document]
JacekClearcode: No example provided.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,842 @@
...
+Example of usage::
+
+    _paq.push([ function() {
+        var customDimension = this.getCustomDimension(1);
+        console.log(customDimension);
+    }]);
+
+Content Tracking
+^^^^^^^^^^^^^^^^
+
```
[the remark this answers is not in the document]
JacekClearcode: Done

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,842 @@
...
+        var customVariable = this.getCustomVariable(1, "visit" );
+        console.log(customVariable);
+    }]);
+
+Custom Dimensions
+^^^^^^^^^^^^^^^^^
+
+Tracking Custom Dimension
+`````````````````````````
+If you want to set custom dimension to use it in tracking functions use ``setCustomDimension`` function::
```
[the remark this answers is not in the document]
JacekClearcode: No difference provided, it should be some sort of help center link maybe?

analytics/tracker_object.rst
JacekClearcode: It should be written in uppercase, fixed

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,820 @@
...
+
+    String with product quantity, optional.
+
+.. warning::
+
+    All string parameters should be escaped, without double quote character.
+
+.. warning::
+
+    You must do this each page view to maintain cart status.
```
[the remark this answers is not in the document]
RomanSek: Proposition:
State of the cart is not maintained across the session/visit [which one we use?]. You must add all products after each page view.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,924 @@
...
+
+.. warning::
+
+    This function does not track any data to Tracker. This is helper function for ``trackEcommerceOrder`` or ``trackEcommerceCartUpdate`` function to populate items to send.
```
RomanSek: I would put this warning before param description. I'm also not sure if it has to be warning. Maybe `note` would be enough.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,924 @@
...
+   **Required** Value of the variable limited to 200 characters.
+
+.. data:: scope
+
+   **Required** Scope of the variable, "visit" or "page"
+
+
+.. warning::
+
+    Custom Variable with scope set on "visit" will be saved for visit, you don't need to save it every page.
```
RomanSek: I think this should be a `note` to the `scope` description. I think you can indent it to make it so.
It would be better to specify how both of those values affect availability of the variable (visit - remembered until end of visit/session [we should make a term for that in glossary to avoid confusion], page - remembered until page is changed).

It's also not clear to me if visit and session are the same in this case. Piwik lets admin set visit max length, but I'm not sure if visit variable is deleted when visit Piwik ends. It's probably stored in session cookie. Can we request clarification in TODO?

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,924 @@
...
+
+    _paq.push([ function() {
+        var customDimension = this.getCustomDimension(1);
+        console.log(customDimension);
+    }]);
+
+Content Tracking
+^^^^^^^^^^^^^^^^
+
+Content Tracking is an useful feature when you want to measure performance of your website or app. For example you can check
```
RomanSek: Typo: "an useful" => "a useful" I think, but I would skip such adjectives and just explain what it does:

Content Tracking tracks how many times specific elements were rendered/visible. It can be used to measure if ad placement was visible or if user have seen end of article.

You mentioned there also file download. I thought that isn't part of this module, but if it is - then such example is also OK.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,924 @@
...
+Code::
+
+    _paq.push(["trackVisibleContentImpressions", checkOnScroll, timeIntervalInMs]);
+
+.. data:: checkOnScroll
+
+    **Required** If set to true it will invoke this function to track new visible content impressions on scroll event.
+    (It won't detect content blocks placed in a scrollable element)
+
+.. data:: timeIntervalInMs
```
RomanSek: I don't like including time unit in the name. It should be explained in description. Name should just convey intention. Maybe `watchInterval` or `interval`?

analytics/js_tracking_api.rst
```diff
@@ -28,10 +30,12 @@ Installing tracking code via Tag Manager
 
 Installing tracking code via code snippet.
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-To install tracking code you must paste it into html file and provide some required settings:
+Installation via snippet should be done only if Tag Manager is not available or you want to track on multiple domains / subdomains.
```
RomanSek: Grammar: "want to track multiple domains / subdomains"

analytics/js_tracking_api.rst
```diff
@@ -28,10 +30,12 @@ Installing tracking code via Tag Manager
...
 
- * ``$PIWIK_URL`` - location of the piwik.js file
- * ``$IDSITE`` - id of website that should be tracked
+This code should be added just before you want to use the API. Additionally snippet has to be configured this way:
+
+ * String ``//{$PIWIK_URL}/`` should be replaced with location of the piwik.js file
```
RomanSek: Proposition: Use same values as in AM API instructions and snippet:

```
* String ``XXX-XXX-XXX-XXX-XXX`` should be replaces with :term:`app ID` (e.g. ``efcd98a5-335b-48b0-ab17-bf43f1c542be``).
* String ``ppms.example.com`` should be replaced with your PPMS domain name.
```

And in snippet:
```
var u="//ppms.example.com/";
_paq.push(["setTrackerUrl", u+"piwik.php"]);
_paq.push(["setSiteId", "XXX-XXX-XXX-XXX-XXX"]);
```

This should make it clear and will be consistent with other docs.

analytics/js_tracking_api.rst
```diff
@@ -144,41 +148,42 @@ Ecommerce tracking
...
 Adding Ecommerce item
 `````````````````````
-To add ecommerce item (for example to track changes in users cart using ``trackEcommerceCartUpdate``) user ``addEcommerceItem`` function::
+To add ecommerce item (for example to track changes in users cart using ``trackEcommerceCartUpdate``) use ``addEcommerceItem`` function::
 
     _paq.push(["addEcommerceItem", productSKU, productName, productCategory, productPrice, productQuantity]);
 
+.. note::
+
+    This function does not track any data to Tracker. This is helper function for :ref:`trackEcommerceOrder <Tracking Ecommerce order>`   or :ref:`trackEcommerceCartUpdate <Updating cart>` function to populate items to send.
```
RomanSek: Proposition:

This function does not send any data to the analytics (tracker, server?). It only prepares Ecommerce cart/order state to be send with :ref:`trackEcommerceOrder <Tracking Ecommerce order>` or :ref:`trackEcommerceCartUpdate <Updating cart>`.

analytics/js_tracking_api.rst
```diff
@@ -435,16 +437,16 @@ To track all visible content impressions you can use ``trackVisibleContentImpres
...
 
-    **Required** If set to true it will invoke this function to track new visible content impressions on scroll event.
+    **Required** ``boolean`` If set to true it will invoke this function to track new visible content impressions on scroll event.
     (It won't detect content blocks placed in a scrollable element)
 
-.. data:: timeIntervalInMs
+.. data:: watchInterval
 
-    **Optional** If set it will invoke this function to track new visible content impressions on every X miliseconds. By default it is set to 750ms.
+    **Optional** ``number`` If set it will invoke this function to track new visible content impressions on every X miliseconds. By default it is set to 750ms.
```
RomanSek: This phrasing suggests that this value can be not set. Is it possible to disable interval checks and base it only on scroll? What value should be send to do that?

analytics/js_tracking_api.rst
```diff
@@ -442,8 +442,9 @@ Code::
 
 .. data:: checkOnScroll
 
-    **Required** ``boolean`` If set to true it will invoke this function to track new visible content impressions on scroll event.
-    (It won't detect content blocks placed in a scrollable element)
+    **Required** ``boolean`` If ``true`` it will check new visible content impressions on scroll event. Default: ``true``.
+    .. note::
+    It won't detect content blocks placed in a scrollable element.
```
RomanSek: There's an error here. You need to indent text of the note or write it inline:
`.. note:: text`

```
.. note::
    test
```

conf.py
```diff
@@ -62,7 +62,7 @@
 
 # The master toctree document.
 master_doc = 'index'
-autosectionlabel_prefix_document = True
+autosectionlabel_prefix_document = True;
```
RomanSek: wat?

analytics/js_tracking_api.rst
```diff
@@ -46,9 +46,9 @@ This code should be added just before you want to use the API. Additionally snip
       _paq.push(["trackPageView"]);
       _paq.push(["enableLinkTracking"]);
       (function() {
-        var u="//{$PIWIK_URL}/";
+        var u="ppms.example.com";
```
RomanSek: I think slashes have to stay to build URLs. So it should be `"//ppms.example.com/"`.

analytics/js_tracking_api.rst
```diff
@@ -154,7 +154,8 @@ To add ecommerce item (for example to track changes in users cart using ``trackE
 
 .. note::
 
-    This function does not track any data to Tracker. This is helper function for :ref:`trackEcommerceOrder <Tracking Ecommerce order>`   or :ref:`trackEcommerceCartUpdate <Updating cart>` function to populate items to send.
+    This function does not send any data to the analytics (tracker, server?). It only prepares Ecommerce cart/order state to be send with :ref:trackEcommerceOrder <Tracking Ecommerce order> or :ref:trackEcommerceCartUpdate <Updating cart>.
```
RomanSek: There's a 
```
:term:`analytics`
```
 in glossary that can be used to clear it up.

analytics/tracker_object.rst
```diff
@@ -0,0 +1,543 @@
...
+.. function:: storeCustomVariablesInCookie();
+
+Function will enable storing ``"visit"`` type custom variables in a first party cookie.
+That will enable getting them via ``getCustomVariable`` function.
+
+
+Custom dimensions
+-----------------
+
+.. function:: setCustomDimension(customDimensionId, customDimensionValue);
```
wronan: May be worth mentioning, that this function doesn't call backend, just sets CD to be sent when methods like `trackPageView` and similar (conversions, etc...) are used

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,829 @@
...
+Adding extension to default extensions
+++++++++++++++++++++++++++++++++++++++
+
+You can add extension to default extensions list using ``addDownloadExtensions`` function::
+
+    _paq.push(["addDownloadExtensions", extensions]);
+
+.. data:: extensions
+
+    **Required** ``string`` Extensions separated by ``|`` for example ``"mhj|docx"``
```
wronan: String of elements divided by the pipe - that's OK but this parameter also accepts array of strings as a value.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,829 @@
...
+Replacing default extensions list
++++++++++++++++++++++++++++++++++
+
+Default extensions list can be overwritten using ``setDownloadExtensions`` function::
+
+    _paq.push(["setDownloadExtensions", extensions]);
+
+.. data:: extensions
+
+    **Required** ``string`` Extensions separated by ``|`` for example ``"7z|apk|mp4"``
```
wronan: Same here - array of strings is also a acceptable value.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,829 @@
...
+
+You can disable download and outlink tracking for links using CSS classes, simply add ``piwik_ignore`` css class.
+
+To disable using CSS class you can use ``setIgnoreClassess`` function::
+
+    _paq.push(["setIgnoreClasses", className);
+
+.. data:: className
+
+    **Required** ``string`` Css class name that will be ignored
```
wronan: This parameter also accepts an array of strings as possible value.

analytics/js_tracking_api.rst
```diff
@@ -0,0 +1,829 @@
...
+
+Tracking internal searches
+``````````````````````````
+To track search requests on your site use ``trackSiteSearch`` function::
+
+    _paq.push(["trackSiteSearch", keyword, category, searchCount, dimensions]);
+
+.. data:: keyword
+
+    **Required** ``string`` Keyword that was searched
```
wronan: Only first parameter is required, the rest is optional.

analytics/snippet_tracking.rst
```diff
@@ -0,0 +1,123 @@
...
+.. code-block:: html
+
+    <!-- Piwik -->
+    <script type="text/javascript">
+      var _paq = _paq || [];
+      _paq.push(["trackPageView"]);
+      _paq.push(["enableLinkTracking"]);
+      (function() {
+        var u="//ppms.example.com/";
+        _paq.push(["setTrackerUrl", u+"piwik.php"]);
```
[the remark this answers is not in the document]
quba: help, piwik.js and piwik.php for BC, ppms.php and ppms.js starting from PPMS 5.5.1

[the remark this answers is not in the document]
JacekClearcode: Do we should write a note below, with info that for older versions piwik is still supported?

analytics/snippet_tracking.rst
```diff
@@ -0,0 +1,123 @@
...
+
+
+Installing tracking code via code snippet.
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Installation via snippet should be done only if Tag Manager is not available or you want to track multiple domains / subdomains.
+
+This code should be added near the top of the ``<head>`` tag and before any other script or CSS tags. Additionally
+snippet has to be configured this way:
+
+    * String ``XXX-XXX-XXX-XXX-XXX`` should be replaces with :term:`app ID` (e.g. ``efcd98a5-335b-48b0-ab17-bf43f1c542be``).
```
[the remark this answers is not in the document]
JacekClearcode: Links to glossary that Roman wrote in another PR

analytics/snippet_tracking.rst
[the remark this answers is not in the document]
wronan: It's quite around: for older versions only piwik.* is supported, for newers (5.5.1+) we recommend ppms.* but piwik.* will also work as a backward compatibility.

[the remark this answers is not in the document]
JacekClearcode: Hmm so we could leave it as it is and write a  note about suggesting using ppms when using 5.5.1+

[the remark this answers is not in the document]
JacekClearcode: Or another way around.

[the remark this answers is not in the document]
wronan: I don't know what what's the best practice here. We should probably version the docs. Anyway, it's not like 'name doesn't matter since both ppms.js and piwik.js will work' - we renamed it to ppms.* because it was recognized as Matomo by some software.
We should at least remember to update the doc in near future or add some simple note now.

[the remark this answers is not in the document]
JacekClearcode: The docs will be versioned. It’s just an initial release. 😄

PR #11, opened 2018-04-03
PPCDEV-6926 Document review and suggested changes

sebastianpiskorski:


analytics/index.yaml
```diff
@@ -101,47 +102,46 @@ paths:
...
         - name: _cvar
           in: query
-          #TODO: I don't understand this parameter or it's format. HALP!
-          #TODO: Confirm serialization format (it's supposed to be Array JSON, but example suggests Array-like Object JSON).
           #TODO: Confirm description of item format.
-          #TODO: Add limits of item values (e.g. max length).
-          #TODO: Link to a page describing custom variables.
+          #TODO: Link to a definition of scopes.
           description: |
             <a href="https://help.piwik.pro/analytics/custom-variables/">Custom variables</a> list set on the ``visit``
```
RomanSek: To make it clear I would remove `list` from this sentence:
`Custom variables set on the visit scope`.

analytics/index.yaml
```diff
@@ -101,47 +103,46 @@ paths:
...
             scope.
 
+            Custom variable param is a JSON object where each key is an ID of a custom variable and each key contains
```
RomanSek: I think we can skip name of parameter here. This is continuation of description and we already mentioned name of the parametr in first sentence. Value definition is also separated from definition of the object. Maybe we can combine everything this way: 
```
Format: Object serialized with JSON:

  - key - (`string`) Custom variable ID
  - value - Array with:
    - (`string`) Custom variable name, max length: 200 characters
    - (`string`) Custom variable value, max length: 200 characters
```

`cvar` description should be also updated to keep them unified.

analytics/index.yaml
```diff
@@ -254,9 +252,10 @@ paths:
...
         - name: cid
           in: query
-          #TODO: What's the difference between _uid, cid and uid parameters? Are some of those deprecated?
           description: |
-            Visitor ID.
+            This parameter overwrites visitors `config_id`. For further explanation please read following article:
+            <a href="https://help.piwik.pro/analytics/recognizing-visitors/">Recognizing Visitors</a> - the
+            `Fingerprint` section.
 
             It allows to use external application identifier of a Visitor instead of default identifier generated by
```
RomanSek: This sentence is not true if I understand your description. I would remove it.

analytics/index.yaml
```diff
@@ -254,9 +252,10 @@ paths:
           example: PhilipFry
         - name: cid
           in: query
-          #TODO: What's the difference between _uid, cid and uid parameters? Are some of those deprecated?
           description: |
-            Visitor ID.
+            This parameter overwrites visitors `config_id`. For further explanation please read following article:
```
RomanSek: I'm trying to give these information parameter description (in this sequence):

* full name of the parameter used by other parts of the documentation. Ideally user should be able to refer to concept represented by parameter using it and search for it in docs or help site. Example: `Configuration ID` or `Fingerprint ID`.
* Short description of the concept. It doesn't have to be extensive, just a summary with a link to full article will do. Example: `semi-unique hash generated for Visitor's client (based on configuration and installed plugins)`.
* Format of the parameter (is it's too complicated to describe with OpenAPI options) 

Can you reformat description to add these information's?

PR #24, opened 2018-06-19
Added various description fixes PPROF-1163

RomanSek:


audience_manager/authorized_api/paths/audiences_list.yaml
```diff
@@ -25,19 +25,22 @@ get:
...
     description: >
-      Sorting order for the results. Include <code>-</code> prefix in value
-      to get a descending order.
+      Sorting order for the results. Values with `-` prefix return results in descending order.
     default: name
     enum:
     - name
+    - -name
     - count
+    - -count
```
nfpp: How it works with two `- -`?

RomanSek: It is treating it as part of the string, so we have `"count", "-count", ...`

PR #25, opened 2018-06-19
Description fixes PPTECHDOC-37

RomanSek:


analytics/js_tracking_api.rst
```diff
@@ -732,18 +732,18 @@ Example of usage
...
 
-Setting Download delay
-++++++++++++++++++++++
+Setting link tracking delay
++++++++++++++++++++++++++++
 
-After each outbound link there is a small time frame after which the file will download that will
-ensure there is enough time to track downloads.
-That time frame is set by default to 500ms. To modify it you can use ``setLinkTrackingTimer`` function::
+After each outbound/download link click there is introduced a small delay after which the browser
```
wronan: I'm not an expert but shouldn't it rather be "... there is a small delay introduced ..." - somehow sounds better to me.

analytics/js_tracking_api.rst
```diff
@@ -732,18 +732,18 @@ Example of usage
...
+will navigate to new URL. This ensures there is enough time to track link interactions.
+That time frame is set by default to 500ms. To modify it you can use the ``setLinkTrackingTimer`` function::
 
     _paq.push(["setLinkTrackingTimer", time]);
 
 .. describe:: time
 
-    **Required** ``number`` Time in ms between user interaction and downloading file.
+    **Required** ``number`` Time in ms between user action (click) and changing a page / downloading a file.
```
wronan: It's more like "changing a website" then "changing a page" - it doesn't matter much to me but average user may think that this delay is forced after each page view (it isn't).

analytics/tracker_object.rst
```diff
@@ -408,7 +416,7 @@ Tracker Configuration
 
 .. function:: setDomains(domains)
 
-    The function that will set an array of domains to be treated as local. Wildcards, dots are supported for subdomains.
+    The function that will set an array of domains to be treated as local. Sub-domain wildcards are supported (e.g. ``*.example.com``).
```
wronan: :+1:

analytics/js_tracking_api.rst
wronan: You get the idea? Let's say you have a link to some PDF file - this is considered outbound link because once user get's there, you don't know what he/she/it does next. Same goes for external websites (let's say you have a link to wp.pl on you website) - once user clicks on such link he/she/it can't be tracked anymore. These are two special situations when we have to postpone the action propagation on order to give proper amount of time for that one last tracking request to complete.

PR #42, opened 2018-07-10
feature/PPCDEV-7517 public oauth guide introduction, public docs for: Apps, Users, Access Control

plks:
* public oauth guide introduction
* public docs for: Apps, Users, Access Control

### DIFF
--- a/platform/public_api/access_control/index.json
+++ b/platform/public_api/access_control/index.json
@@ -2,7 +2,7 @@
   "swagger": "2.0",
   "info": {
     "title": "Access Control Api",
-    "version": "2.0"
+    "version": "auto"
   },
   "parameters": {
     "Entity_Id": {
@@ -34,7 +34,7 @@
     "JsonApi_Id": {
       "type": "string",
       "format": "uuid",
-      "description": "UUIDv4 string",
+      "description": "Resource ID",
       "example": "985ed9d7-b652-47b8-b9cf-c17d62735261"
     },
     "JsonApi_Type": {
--- a/platform/public_api/access_control/index.json
+++ b/platform/public_api/access_control/index.json
@@ -1,0 +1,470 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Access Control Api",
+    "version": "2.0"
+  },
+  "parameters": {
+    "Entity_Id": {
+      "in": "path",
+      "name": "entityId",
+      "type": "string",
+      "format": "uuid",
+      "required": true,
+      "description": "UUIDv4 identifier of given entity"
+    },
+    "Entity": {
+      "in": "path",
+      "name": "entity",
+      "description": "entity",
+      "type": "string",
+      "required": true,
+      "enum": [
+        "ppms/app",
+        "ppms/user"
+      ]
+    }
+  },
+  "tags": [
+    {
+      "name": "AccessControl v2"
+    }
+  ],
+  "definitions": {
+    "JsonApi_Id": {
+      "type": "string",
+      "format": "uuid",
+      "description": "UUIDv4 string",
+      "example": "985ed9d7-b652-47b8-b9cf-c17d62735261"
+    },
+    "JsonApi_Type": {
+      "type": "string",
+      "description": "Resource type, unique across whole PPMS",
+      "example": "ppms/user"
+    },
+    "JsonApi_Resource_Object": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "$ref": "#/definitions/JsonApi_Type"
+        },
+        "id": {
+          "$ref": "#/definitions/JsonApi_Id"
+        },
+        "attributes": {
+          "type": "object"
+        }
+      }
+    },
+    "Entity_Actions_Get_Response": {
+      "type": "object",
+      "properties": {
+        "data": {
+          "allOf": [
+            {
+              "$ref": "#/definitions/JsonApi_Resource_Object"
+            },
+            {
+              "properties": {
+                "attributes": {
+                  "type": "object",
+                  "properties": {
+                    "actions": {
+                      "type": "array",
+                      "items": {
+                        "type": "string"
+                      },
+                      "description": "List of available actions for given entity"
+                    }
+                  }
+                }
+              }
+            }
+          ]
+        }
+      },
+      "example": {
+        "data": {
+          "type": "ppms/permission/entity/actions",
+          "id": "50fca71a-e4a7-4480-a612-a9ffbdf0935f",
+          "attributes": {
+            "actions": [
+              "view",
+              "edit",
+              "publish",
+              "delete",
+              "manage"
+            ]
+          }
+        }
+      }
+    },
+    "Entity_Actions_Get_Users_Response": {
+      "type": "object",
+      "properties": {
+        "meta": {
+          "type": "object",
+          "description": "Meta information",
+          "properties": {
+            "total": {
+              "type": "integer",
+              "description": "Number of records found"
+            }
+          }
+        },
+        "data": {
+          "type": "object",
+          "allOf": [
+            {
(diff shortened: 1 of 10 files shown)


platform/public_api_guide.md
```diff
@@ -0,0 +1,59 @@
+Public API guide
+========================
+
+This page describes how to access Piwik PRO API which uses 
+[client credentials](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)
+Oauth grant type for obtaining user token.
```
glothriel: We should probably use official notation - `OAuth` -> https://en.wikipedia.org/wiki/OAuth

platform/public_api_guide.md
```diff
@@ -0,0 +1,59 @@
...
+
+This page describes how to access Piwik PRO API which uses 
+[client credentials](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)
+Oauth grant type for obtaining user token. 
+All data is sent and received as JSON and it uses [JSON API](http://jsonapi.org/) specification.
+
+## Obtaining token
+
+If you want to access API for the first time you need to generate your 
+API Credentials which then allows you to request for a token that is used for authentication during communication with public API.
```
glothriel: credentials should probably lowercase.

platform/public_api_guide.md
```diff
@@ -0,0 +1,59 @@
...
+Having generated your API Credentials, now you are ready for creating access token that will be used in communication with API.
+To do so, send POST request 
+```
+POST /auth/token
+```
+
+Curl example:
+
+```
+curl 'https://<domain>/auth/token' -H "Content-Type: application/json" --data '{
```
glothriel: Is this endpoint also documented in OpenAPI?

RomanSek:
You should merge it to branch containing first version of PPMS that will have this functionality. Which one is it? 5.5, 6.0, 6.1?

JacekClearcode:
MD files are supported, but we strongly recommend using RST files. http://ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/

I don't mean to urge you to rewrite it all to `.rst` but you can use it in other documentations.

platform/public_api_guide.md
JacekClearcode: Can't find it there.

jzyinq:
I see that `version` tag is overwritten to `6.0` - remember that this docs are related to `6.1` version so responding to @RomanSek comment there is no need for backporting these docs.

JacekClearcode:
@jzyinq But still this PR must be merged to version that will have this functionality (or already has it) merging to master should be prohibited.

jzyinq:
@JacekClearcode I don't see any `6.1` branch so in this context this is `latest` version of docs.

Do you have some kind of `develop` branch like in PPMB repo from which we need to branch out ?

JacekClearcode:
@jzyinq  you can branch out from master. Then when all tasks from 6.1 will be completed that branch will be merged to master without deleting itself. That way latest will point to the same thing as 6.1 and when new version will come out 6.1 still will be available to the user in the RTD menu. ;)

platform/public_api_guide.md
plks: no, this endpoint is not documented via OpenAPI. We wanted to describe it via public api guide because it's some kind of special and not JsonApi complaint. Also we have other endpoints in `OAuth` service, but docs are unfortunately not ready yet.

platform/public_api/access_control/index.json
```diff
@@ -0,0 +1,470 @@
...
+  "tags": [
+    {
+      "name": "AccessControl v2"
+    }
+  ],
+  "definitions": {
+    "JsonApi_Id": {
+      "type": "string",
+      "format": "uuid",
+      "description": "UUIDv4 string",
```
RomanSek: It describes format instead of what this param is used for / what it is. For format description use `format` attribute.

platform/public_api/apps/index.json
```diff
@@ -0,0 +1,744 @@
...
+    },
+    "App_AddedAt": {
+      "description": "When app was added",
+      "type": "string",
+      "example": 1506570461
+    },
+    "JsonApi_Id": {
+      "type": "string",
+      "format": "uuid",
+      "description": "UUIDv4 string",
```
RomanSek: Description specifies format here instead of explaining meaning of the param.

platform/public_api/apps/index.json
```diff
@@ -0,0 +1,744 @@
...
+        },
+        "excludedIps": {
+          "description": "Excluded ips for an app",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "example": [
+            "123.43.32.10",
+            "8.8.8.8"
```
RomanSek: It may be safer to use private IPs in examples.

platform/public_api/apps/index.json
```diff
@@ -0,0 +1,744 @@
...
+        },
+        "privacyCookieDomains": {
+          "description": "List of privacy cookie domains for an app (works if enabled)",
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "example": [
+            "*.piwik.pro",
+            "careers.piwik.pro"
```
RomanSek: I tried to stick to `example.com` domain and subdomains in examples.

platform/public_api/users/index.json
```diff
@@ -0,0 +1,402 @@
...
+          "400": {
+            "$ref": "#/responses/Bad_Request"
+          },
+          "404": {
+            "$ref": "#/responses/Not_Found"
+          }
+        }
+      }
+    },
+    "/api/users/v2/{uuid}/change-role": {
```
RomanSek: Not a great description of parameter. It tells me format of parameter, but not what it is used for. I also see that it's definition doesn't have any description, so only way to identify what it does is context.

platform/public_api_guide.md
RomanSek: Users will assume that every endpoint is described to together. Splitting endpoint description like that may not be worth it if it'll cause confusion.

jzyinq: Our whole documentation lacks information about which endpoints are protected with OAuth/JWT (tag `security`) and which aren't so at the moment user is still obliged to pass through this section if he wants to use our API.

Im not arguing that this will not be better but I think that we could fix this in next iteration of this docs when we have some feedback from end users.

platform/public_api/access_control/index.json
```diff
@@ -0,0 +1,470 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "title": "Access Control Api",
+    "version": "2.0"
```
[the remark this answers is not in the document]
plks: we are aware, that our microservices starts from API v2 version and docs are in different versioning system (v6.1). We assume that docs will be changing more frequently than api versioning and do not consider it as issue.

platform/public_api/access_control/index.json
plks: actually its a definition for JsonAPI object, so sometimes its a `entity ID`, sometimes `User ID`. @jzyinq do you agree?

platform/public_api/apps/index.json
plks: here, context is always App so now its `App ID`

platform/public_api/access_control/index.json
jzyinq: That's correct. Since it's generally returns same id that was queries maybe we can name this something like `Id of requested resource` ?

platform/public_api/access_control/index.json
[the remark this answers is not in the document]
RomanSek: ok

platform/public_api/access_control/index.json
plks: `Resource ID`

PR #51, opened 2018-07-23
PPROF-1132: Added documentation for janus-protected endpoints in AM, …

glothriel:
…cherry picked e71f7ac 0c22b76 dc138a6 c388d8e e706c47.

Well, I managed to break previous PR by merging master and then unsuccesfully trying to remove extra commits that should not have been merged onto this version.

Here is new one, with cherry-picked commits from the previous PR and my changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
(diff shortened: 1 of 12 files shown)


glothriel:
This is for `6.1.1`, should we merge it to `6.1` or create new branch?

JacekClearcode:
We are using SemVer standard without PATCH version (ex. 1.0) so i think this can go to `6.1`

PR #55, opened 2018-07-31
PPCDEV-7059 Apps api docs

mrfijas:
* gdpr manual mode
* cookie expiration period
* small fixes of existing docs

### DIFF
--- a/platform/public_api/apps/index.json
+++ b/platform/public_api/apps/index.json
@@ -327,10 +327,20 @@
           "type": "boolean",
           "example": true
         },
+        "gdprUserModeEnabled": {
+          "description": "Gdpr User Mode enabled or disabled for an app",
+          "type": "boolean",
+          "example": false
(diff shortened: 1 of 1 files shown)


RomanSek:
I want to confirm that these changes were introduced in version 6.2 because they look to me like documentation corrections. If these are in fact only corrections - we should merge them into first version that introduced there elements. It looks like we added changed file in v6.1.

mrfijas:
@RomanSek you're right only two fields were added in `6.1`, rest are small fixes that we've found in the meantime.. so I'll add another "backport PR" to `6.1` with corresponding changes, thanks for good eye!

RomanSek:
@mrfijas I rebased it to corrected version of 6.2. Let me know if changes match what you intended.

mrfijas:
@RomanSek yup, it's intended change, thanks

PR #57, opened 2018-07-31
Describing new parameter for the tracking API and new method for the …

wronan:
…JS tracker

### DIFF
--- a/analytics/index.yaml
+++ b/analytics/index.yaml
@@ -652,6 +652,17 @@ paths:
           minimium: -180
           maximum: 180
           example: -74.0023705
+        - name: uia
+          in: query
+          description: |
+            Allows to track user anonymously.
+
+            - ``1`` - all IP bytes will be masked (0.0.0.0), no GeoIP data will be stored and visits won't be matched by the fingerprinting process
+            - ``0`` - all available visitor data will be enriched if anonymized previously
+          type: integer
+          minimium: 0
+          maximum: 1
+          example: 1
       responses:
         200:
           description: OK
--- a/analytics/js_tracking_api.rst
+++ b/analytics/js_tracking_api.rst
@@ -839,3 +839,19 @@ Example of usage::
 
     _paq.push(["trackSiteSearch", "test", false, 20]);
 
+
+
+Tracking user anonymously
+``````````````````````````
+To track visitor anonymously (without consent) use the ``setUserIsAnonymous`` function::
+
+    _paq.push(["setUserIsAnonymous", isAnonymous]);
+
+.. describe:: isAnonymous
+
+    **Required** ``boolean`` Flag that sets anonymous tracking on and off
+
+Example of usage::
+
+    _paq.push(["setUserIsAnonymous", true]);
+


analytics/js_tracking_api.rst
```diff
@@ -839,3 +839,19 @@ Example of usage::
...
+
+    _paq.push(["setUserIsAnonymous", isAnonymous]);
+
+.. describe:: isAnonymous
+
+    **Required** ``integer/boolean`` Flag that sets anonymous tracking on and off (true/false or 1/0)
+
+Example of usage::
+
+    _paq.push(["trackSiteSearch", true]);
```
azarecki: Please change trackSiteSearch to setUserIsAnonymous

analytics/index.yaml
```diff
@@ -652,6 +652,18 @@ paths:
...
           maximum: 180
           example: -74.0023705
+        - name: uia
+          in: query
+          description: |
+            Allows to track user anonymously.
+
+            When set to 1 all IP bytes will be masked (0.0.0.0), no GeoIP data will be stored
+            and visits won't be matched by the fingerprinting process. When set back to 0 (or ommited)
+            all available visitor data will be enriched.
```
RomanSek: When values were enumerated I was trying to describe them in a list. Check how `send_image` param is described in this file for example.

OpenAPI has also `default` attribute for params.

analytics/js_tracking_api.rst
```diff
@@ -839,3 +839,19 @@ Example of usage::
...
+
+Tracking user anonymously
+``````````````````````````
+To track visitor anonymously (without consent) use the ``setUserIsAnonymous`` function::
+
+    _paq.push(["setUserIsAnonymous", isAnonymous]);
+
+.. describe:: isAnonymous
+
+    **Required** ``integer/boolean`` Flag that sets anonymous tracking on and off (true/false or 1/0)
```
RomanSek: I think that alternative types are usually separated by `|` character (e.g. in [JSDocs](http://usejsdoc.org/tags-type.html)), but I see that this character is also used in our docs.

Try to format JS values in descriptions. Check how it's done in `checkOnScroll` description in this file.

analytics/js_tracking_api.rst
```diff
@@ -839,3 +839,19 @@ Example of usage::
...
+
+Example of usage::
+
+    _paq.push(["trackSiteSearch", true]);
+
```
RomanSek: I see that RST format has a [versionadded](http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionadded) directive to notify user when element was introduced. Maybe we should use that at the end of description? We didn't use it yet, but you can find `deprecated` directive in this file that conveys similar information.

analytics/js_tracking_api.rst
wronan: F***...

analytics/js_tracking_api.rst
wronan: Hmm... setEcommerceView parameters description mislead me.

wronan: I'll keep it simple and stick with boolean.

PR #60, opened 2018-08-01
PPCDEV-7059 Apps - small api docs fixes

mrfijas:


RomanSek:
@mrfijas can we merge it? Were these changes already checked meteorically in #55?

PR #62, opened 2018-08-01
[INTERNAL] Changed texts on title and copyright

RomanSek:
Looks like we didn't propagate these changes to future versions. Can you confirm that all configs should be updated?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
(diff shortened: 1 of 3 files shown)


RomanSek:
Ok, I see that these changes are on all branches, but each branch has in in different commit:
5.5 - c71e6e01c05d23ca922f19849cb587fc9ea720c4
6.0 - 408072b83d46d730c472ca98c40f99d5ac988073
6.1 - 4e4eb7bbc3cecd2d4ea14f946520226ada86b845
That's why the show up again. We'll need to decide what we'll do with this.

PR #71, opened 2018-08-03
app type

chriskicker:


JacekClearcode:
@chriskicker Is this all new things? I think some of these changes are just documentation corrections of things already introduced in an earlier version.

chriskicker:
@JacekClearcode there are no significant changes (no bugfixes, change params order and real life examples introduced only), so imo it could go to 7.0 only

RomanSek:
It would be better to keep docs as consistent as possible. If we're adding new info to old parts - we should try to keep them consistent across doc versions.

RomanSek:
@chriskicker I think that creating another PR with updates should be rather easy. Just branch out from older version and patch it with these changes (removing new functionality). After we'll merge them upstream - we will rebase this PR to remove repeated changes and that should be enough.

RomanSek:
Rebased and merged as https://github.com/PiwikPRO/PPMS-PublicDocs/pull/75

PR #108, opened 2018-09-19
Feature/ppcdev 8176 step by step guide

plks:
step by step guide -> curl + Postman example

### DIFF
--- a/platform/authorized_api/users/index.json
+++ b/platform/authorized_api/users/index.json
@@ -13,6 +13,9 @@
       "name": "Users v2"
     }
   ],
+  "consumes": [
+    "application/vnd.api+json"
+  ],
   "definitions": {
     "User_Role": {
       "type": "string",
@@ -247,9 +250,6 @@
         ],
         "summary": "Create user",
         "operationId": "api_user_add_v2",
-        "consumes": [
-          "application/vnd.api+json"
-        ],
         "produces": [
           "application/vnd.api+json"
         ],
@@ -281,9 +281,6 @@
         ],
         "summary": "Get users list",
         "operationId": "api_users_list_v2",
-        "consumes": [
-          "application/vnd.api+json"
-        ],
         "produces": [
           "application/vnd.api+json"
         ],
@@ -321,9 +318,6 @@
         ],
         "summary": "Get details of an user",
         "operationId": "api_users_user_details_v2",
-        "consumes": [
-          "application/vnd.api+json"
-        ],
         "produces": [
           "application/vnd.api+json"
         ],
@@ -341,9 +335,6 @@
         "operationId": "api_user_delete_v2",
         "tags": [
           "Users v2"
-        ],
-        "consumes": [
-          "application/vnd.api+json"
         ],
         "responses": {
           "204": {
@@ -374,9 +365,6 @@
         ],
         "summary": "Change user role",
         "operationId": "api_users_change_role_v2",
-        "consumes": [
-          "application/vnd.api+json"
-        ],
         "parameters": [
           {
             "name": "User with role",
@@ -407,9 +395,6 @@
         ],
         "summary": "Change my password",
         "operationId": "api_me_change_password_v2",
-        "consumes": [
-          "application/vnd.api+json"
-        ],
         "parameters": [
           {
             "name": "User with new password",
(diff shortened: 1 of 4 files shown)


platform/authorized_api_guide.md
```diff
@@ -20,40 +21,317 @@ API credentials which then allows you to request for a token that is used for au
   Name must contains at least 3 characters.
 * Copy your newly generated `CLIENT ID` and `CLIENT SECRET` because they **won't be available for you after dismissing this window.**
 
-### Create access token
+This credentials will be valid as long as you revoke them in your profile.
```
RomanSek: I have problem with understanding this sentence:
* Linguistic: "This credential ..." or "Those credentials ..." are correct, "This credentials ..." is not.
* Logic: Credentials will be valid if I'll revoke them or as long as I wont revoke them?

platform/authorized_api_guide.md
```diff
@@ -20,40 +21,317 @@ API credentials which then allows you to request for a token that is used for au
...
+Response example:
 ```
-POST /auth/token
+{"token_type":"Bearer","expires_in":86400,"access_token":"<your_access_token>"}
 ```
 
-Curl example:
+Now, you can use obtained `<your_access_token>` for communication with Piwik PRO API.
+Field `expires_in` stands for time (in seconds) for token expiration (TTL).
+As token is a Bearer type, it must be **included in every API call** within header.
```
RomanSek: Maybe change it to: "Since token ..."?

platform/authorized_api_guide.md
```diff
@@ -20,40 +21,317 @@ API credentials which then allows you to request for a token that is used for au
...
+Whatever API call you choose, first remember that you must generate 
+[API credentials](#generate-api-credentials) for obtaining client id and secret.
+
+### API usage example with curl
+
+For sake of this examples, `https://<domain>` is a URL of your PPMS instance (e.g. `https://example.piwik.pro`) and our goal
+is to perform basic operations on an app. We will:
+* create an app
+* get created app
+* update its some fields
```
RomanSek: I don't get it. Should this be "update its fields/attributes"?

platform/authorized_api_guide.md
```diff
@@ -20,40 +21,317 @@ API credentials which then allows you to request for a token that is used for au
...
+* in the end, we will remove given app
```
RomanSek: Just "remove the app" should be enough in this point.

platform/authorized_api_guide.md
```diff
@@ -20,40 +21,317 @@ API credentials which then allows you to request for a token that is used for au
...
+```
+{
+   "data":{
+      "type":"ppms\/app",
+      "id":"b30e538d-4b05-4a75-ae25-7eb565901f38",
+      "attributes":{
+         "name":"AppName",
+         "addedAt":"2018-09-13T12:16:30+00:00",
+         "urls":[
+            "http:\/\/example.com"
```
RomanSek: Is this correct? I don't think JSON escapes slashes, but it shouldn't affect build docs. It's just harder to read in the source.

platform/authorized_api_guide.md
```diff
@@ -20,40 +21,317 @@ API credentials which then allows you to request for a token that is used for au
...
+
+Simply click in Postman: `import -> Import From Link`. Then all of your paths are imported!
+You have to override two things:
+* replace your domain in url
+* add token. Click on `Authorization` tab on chosen API call and then use Bearer Token type. 
+  Paste your token and now you can call API using `SEND` button.
+
+## FAQ
+
+Here you can find most common issues during work with API.
```
RomanSek: Maybe "Here you can find the most common issues encountered during work with the API"?

platform/authorized_api/users/index.json
```diff
@@ -397,6 +417,31 @@
...
+    },
+    "/api/users/v2/me/details": {
+      "get": {
+        "tags": [
+          "Users v2"
+        ],
+        "summary": "Current user details",
+        "operationId": "api_user_details_v2",
+        "produces": [
+          "application/vnd.api+json"
```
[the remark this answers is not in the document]
plks: no, sometimes we return `204` without Content Type sometimes we produce it. It cannot be generalized as `consumes` section

platform/authorized_api_guide.md
plks: yes it is semantically correct, but as you pointed it's not needed

PR #117, opened 2018-11-05
PPCDEV-8489 Added Piwik anonymization JS tracker methods

azarecki:


JacekClearcode:
@azarecki These things are included in 7.0 version only, yes?

azarecki:
From version 7.0 in all versions, should I make additional PR to some branch?

PR #119, opened 2018-11-19
Faeture/ppcdev 8561 new platform endpoints

chriskicker:


RomanSek:
Change destination to correct version branch and tell us if this is indeed just a bugfix. I see that there are added parameters. Were they added in previous version and were not added in documentation?

JacekClearcode:
Right now I don't know if these fixes will be added to `7.0` version, or are being added to another existing version.

chriskicker:
base branch changed, language and data anonymization are new fields

platform/authorized_api/users/index.json
```diff
@@ -19,14 +19,16 @@
   "definitions": {
     "User_Role": {
       "type": "string",
+      "description": "User role",
```
RomanSek: Do you want to add missing descriptions to previous versions?

platform/authorized_api/users/index.json
```diff
@@ -147,8 +158,18 @@
...
                 },
                 "email": {
-                  "type": "string",
-                  "format": "email"
+                  "$ref": "#/definitions/User_Email"
+                },
+                "role": {
+                  "allOf": [
+                    {
+                      "$ref": "#/definitions/User_Role"
```
RomanSek: FYI: Names of definitions are used in documentation:
screenshot from 2018-11-19 16-35-36
It may be good idea to rethink them.

platform/authorized_api/users/index.json
chriskicker: imo good point, @plks will you prepare a PR?

RomanSek: If you want to do that - it'll be better to wait with merge of this branch until these changes will be on target branch. I'll let us rebase this branch against it and remove duplication of description changes in this PR.

platform/authorized_api/users/index.json
plks: thanks, we will talk about that

PR #120, opened 2018-11-21
PPNA-875 added copy api call screenshot

vanzi:


custom_reports/http_api/paths/query.yaml
```diff
@@ -11,6 +11,8 @@ post:
     available in Custom Reports application in each report to get more
     usage examples based on your existing reports.
 
+    ![View API call definition](../../_static/images/copy-api-call.png)
```
RomanSek: It's markdown syntax. It'll work in Redoc, but will not render image in PDF (restructured text image is added this way: `.. image:: images/ball1.gif`). I think that in non-HTML builds it'll leave this text instead.

PR #136, opened 2019-01-09
PPROF-1288 updated attributes api docs

vanzi:


audience_manager/public_api/definitions/attributes_data.yaml
```diff
@@ -40,13 +40,28 @@ properties:
     description: Attributes of the occurred event.
     items:
       type: object
+      required:
+        - name
+        - value
```
RomanSek: It's not required when action is `add`. Maybe we should resign from that default to simplify API?

vanzi: Hidden feature :stuck_out_tongue:

audience_manager/public_api/definitions/attributes_data.yaml
```diff
@@ -40,13 +40,28 @@ properties:
...
+        action:
           type: string
-          description: Value of the attribute.
+          enum:
+            - set
+            - add
+          default: set
+          description: |
+            Action to perform:
+            * `set` – override the existing value with the new value,
```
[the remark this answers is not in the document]
vanzi: Either way is fine for me, you can change it if you want.

PR #147, opened 2019-02-11
PPNA-1250 added metric_filters to custom reports API docs

vanzi:


custom_reports/columns.rst
```diff
@@ -381,7 +381,7 @@ dimensions to metrics or different dimensions.
     +-------------------+-----------------+----------------+-----------+
     |Max                |max              |int, float      |(as source)|
     +-------------------+-----------------+----------------+-----------+
-    |Average            |average          |int, float, bool|(as source)|
+    |Average            |average          |int, float, bool|float      |
```
RomanSek: It looks like correction that should also be propagated to previous version docs. Can you confirm?

custom_reports/http_api/definitions/query_request.yaml
```diff
@@ -99,6 +98,69 @@ properties:
...
+        items:
+          type: object
+          required:
+            - column_id
+            - condition
+          properties:
+            column_id:
+              type: string
+              description: |
+                ID of the column (must be a metric, or a dimension converted
```
RomanSek:
```suggestion
                ID of the column (must be a metric, or a column converted
```

vanzi: Why?

RomanSek: I think we also have metric -> metric conversions. This would be more general description. We can leave previous one if we don't have metric -> dimension conversions.

vanzi: Transformations by definition may only be applied to dimensions (i.e. the raw data).

custom_reports/columns.rst
vanzi: If you feel like it, by all means :joy:

PR #186, opened 2019-07-05
[WIP] PPCDEV-10155 User Group API documentation

mrfijas:


platform/index.rst
```diff
@@ -8,3 +8,4 @@ Platform
   authorized_api/users/users_api
   authorized_api/apps/apps_api
   authorized_api/access_control/access_control_api
+  authorized_api/user_groups/user_groups_api
```
jzyinq: Can we rearrange that index alphabetically? 

obraz

It's small improvement but would look a lot better imo. Maybe with an exception for `Authorized Api Guide` as it's an introduction how to connect with our api. I would leave it at top.

platform/authorized_api/user_groups/v1.yaml
```diff
@@ -0,0 +1,542 @@
...
+          type: string
+          default: name
+          required: false
+          enum:
+            - name
+            - created_at
+            - updated_at
+        - in: query
+          name: limit
+          description: Limits the number of returned user groups in response
```
jzyinq: I remember we talked about generic descriptions for that kind of parameters e.g not `number returned user groups` but `number of entries`. Could we fix it here right or we leave it for another time? 

I think same thing could apply to `offset` as well.

platform/authorized_api/user_groups/v1.yaml
```diff
@@ -0,0 +1,542 @@
...
+    in: query
+    name: search
+    description: Search phrase for searching for users by email or user id
+    type: string
+    required: false
+
+  UsersSort:
+    in: query
+    name: sort
+    description: Sort type that can be reversed by adding dash before type e.g (-email)
```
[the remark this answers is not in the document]
mrfijas: yeah, definitely 👍
I've started fixing it, so I'm marking this PR as `[WIP]` and resurrect it later..
things like will be fixed soon: https://github.com/PiwikPRO/common-PPMS-AccessControl/blob/83cdac096bf148af1d13a4cd4b4b669ae9d6c334/docs/api/public/user-groups/v1.yaml#L65-L71

platform/authorized_api/user_groups/v1.yaml
mrfijas: yeah, it's fixed on my [PPCEV-10201](https://clearcodehq.atlassian.net/browse/PPCDEV-10201): https://github.com/PiwikPRO/common-PPMS-AccessControl/blob/83cdac096bf148af1d13a4cd4b4b669ae9d6c334/docs/api/public/user-groups/v1.yaml#L489-L503

mrfijas:
ok, I'll make another PR after fixing docs in service, no need to [WIP] here..

PR #191, opened 2019-08-29
TM-5098: Create a guide on how to change data layer name

musialp:


tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,56 @@
+.. highlight:: js
+.. default-domain:: js
+
+Custom data layer name
+======================
+
+Introduction
+------------
+The purpose of this article is to describe the steps that need to be performed to set up custom data layer name. The default value for the data layer name is `dataLayer`. Since PPMS version 10.0.0, there is a possibility to change that value.
```
RomanSek: In public docs we should operate only on MAJOR.MINOR versions since bufixes should not have any effect on API. Here it should be `10.0`.

Additionally you can try using [versionadded](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded) or [versionchanged](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) message to make it stand out more.

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,56 @@
...
+    - If you are using both synchronous and asynchronous snippets, it is recommended to use the same data layer name in both. Nonadherence to this rule may cause unexpected and unwanted behavior.
+    - We strongly advise making sure that the chosen name is not used in any other external software present on your website as well as inside your own website's code. Data layer name should be unique and reserved only for use inside the PPMS container. For additional guidelines regarding the data layer naming process, please refer to `Data layer name guidelines`_.
+
+
+Additional snippet migration steps
+---------------
+
+Once you replace your original snippet and want to use the new data layer name, there is one more step that needs to be taken care of. If you are using direct data layer pushes in your code (e.g. to set a `Data layer` variable), you need to make sure to replace all references to `dataLayer` with the newly selected name, e.g (assuming the new name of `customDataLayer`):
+
+    .. function:: dataLayer.push({event: "test-event"})
```
RomanSek: `.. function::` directive is used to describe function purpose and it's parameters.
It looks like here you want to give an example of code. In that case use [code-block](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block) or format used in other examples (text in first line doesn't matter, just make sure to end line with double colon and indent code below it):
```
Example of usage::

    _paq.push(["addDownloadExtensions", "mhj|docx"]);
```

tag_manager/data_layer_name.rst
musialp: I am trying to figure out where to put the `versionadded` directive for it to make sense. Two options that seem to work is the very top of the file or the start of the introduction section. Do you have any other ideas?

RomanSek: I would try to add it as a extension to this description. See how it looks (exact text may need consideration):
```
...
The default value for the data layer name is `dataLayer`.

.. versionadded:: 10.0
    It is now possible to change that value.
```

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Custom data layer name
+======================
+
+Introduction
+------------
+The data layer is a global JavaScript object, that can be used to pass information form the website to PPMS container. The default value for the data layer name is `dataLayer`.
+
+    .. versionadded:: 10.0
+        The name of the object is customizable. The purpose of this article is to describe the steps that need to be performed to set up custom data layer name.
```
RomanSek: Ah, now I get it. So whole article describes how to use new feature (custom dataLayer names). In that case you were right. This `versionadded` should be at the start of the file. It will look like this:
2019-09-03T143628

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+
+    #. for the synchronous snippet, you need to input your desired name into the highlighted variable
+
+        .. image:: ../_static/images/sync-container.png
+            :alt: Container code for synchronous tags
+
+#. At this point, you can embed the snippet on your website. If you are replacing an existing snippet, make sure to take a look at the `Additional snippet migration steps`_ section for instructions on how to ensure full compatibility.
+
+.. note::
+    - The snippet code was slightly changed compared to PPMS versions pre 10.0, to accommodate for the possibility of changing the data layer name. The new snippet, released alongside version 10.0 is backward compatible, however, to use the functionality described in this document, whole snippet code needs to be replaced on your website.
```
RomanSek: I would drop `slightly`. It changed, that's all that user needs to know.

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+    - If you are using both synchronous and asynchronous snippets, it is recommended to use the same data layer name in both. Nonadherence to this rule may cause unexpected and unwanted behavior.
+    - We strongly advise making sure that the chosen name is not used in any other external software present on your website as well as inside your own website's code. Data layer name should be unique and reserved only for use inside the PPMS container. For additional guidelines regarding the data layer naming process, please refer to `Data layer name guidelines`_.
+
+
+Additional snippet migration steps
```
RomanSek: I would name it as `Snippet migration` or `DataLayer name migration`. It should be clearer when to use it, so user can skip reading description if it doesn't affect him.

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+---------------
+
+Once you replace your original snippet and want to use the new data layer name, there is one more step that needs to be taken care of. If you are using direct data layer pushes in your code (e.g. to set a `Data layer` variable), you need to make sure to replace all references to `dataLayer` with the newly selected name, e.g (assuming the new name of `customDataLayer`)::
```
RomanSek: I would give `dataLayer` and `customDataLayer` code formatting using double quotes to make them stand out more in the whole text. Italic doesn't pop out so much.

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+will become::
+
+    customDataLayer.push({event: "test-event"});
+
+Data layer name guidelines
+--------------------------
+
+To avoid conflicts with your existing code, that could cause unwanted behavior, you need to make sure that the name selected for the data layer object is unique. To ensure that it is not already used by someone, we suggest you run the following command in the console on your website:::
+
+    !Object.keys(window).includes("dataLayer")
```
RomanSek: High level stuff. I think that to be more friendly to older browsers you could use `window.hasOwnProperty('dataLayer')`.

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+
+where `dataLayer` is your chosen name. If this operation returns `true` into the console, then you can safely use this value as the name of your data layer.
```
RomanSek: It may be more readable if you would use variable instead of string in example to identify element that should be changed (e.g. `checkedDataLayerName`).

RomanSek:
FYI: I merged `10.0` branch here. It had docker config to easily view build documentation. See README.

tag_manager/data_layer_name.rst
```diff
@@ -0,0 +1,59 @@
...
+
+    customDataLayer.push({event: "test-event"});
+
+Data layer name guidelines
+--------------------------
+
+To avoid conflicts with your existing code, that could cause unwanted behavior, you need to make sure that the name selected for the data layer object is unique. To ensure that it is not already used by someone, we suggest you run the following command in the console on your website:::
+
+    var dataLayerName = "customDataLayer";
+    !window.hasOwnProperty('dataLayer');
```
RomanSek: I think that you wanted to use here variable defined in line before:
```suggestion
    !window.hasOwnProperty(dataLayerName);
```

musialp: Of course. Thanks for catching that.

PR #211, opened 2019-09-03
PROD-1360 Fixed wrong format of Piwik E-Commerce HTTP API parameter

azarecki:
Change formatting of ec_items parameter

### DIFF
--- a/analytics/index.yaml
+++ b/analytics/index.yaml
@@ -319,34 +319,33 @@ paths:
           example: order-123-customer-321
         - name: ec_items
           in: query
-          #TODO: Confirm format in example is valid.
-          #TODO: Confirm that all attributes of item are required.
           #TODO: Add limits to the item attributes (e.g. string max length).
           description: |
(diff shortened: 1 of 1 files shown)


analytics/index.yaml
```diff
@@ -319,34 +319,33 @@ paths:
...
-              - name: `string` Name of item.
-              - category: `string` Category of item.
-              - price: `number` Price of item.
+              - sku: `string` (required) Stock keeping unit.
+              - name: `string` (optional) Name of item.
+              - category: `string` (optional) Category of item.
+              - price: `number` (optional) Price of item.
                 Currency of this value does not matter, but only one should be used by the application (e.g. USD).
-              - quantity: `integer` Quantity of item.
+              - quantity: `integer` (optional, default to 1) Quantity of item.
```
ZielinskiLukaszPiwik: I think we should use `price of an item` or `item's price` instead of `price of item`.

PR #215, opened 2019-09-05
Feature/pptechdoc 56 testing new openapi

JacekClearcode:
Introducing new` Redoc 2.0.0-rc.14` with support for` OpenApi 3.0`, rewritten in React.

I added some css hacks to force it to look better while embedded in RTD.

Sadly I was forced to remove `sphinxcontrib-openapi` as it was breaking the build when it was trying to process files in `OpenApi 3.0` section with downloads are hidden (HTML and EPUB would also be affected)


RomanSek:
Can we fix this style?
2019-09-11T174223

RomanSek:
It may be an issue with docker build. When I build it locally (without docker) it worked a lot better. :+1: 
2019-09-11T175127

RomanSek:
We'll have to decide what do we do with downloads section (https://readthedocs.org/projects/piwik-pro-pdfd/downloads/). I know that we planned to remove PDF and ePub, but HTMLZip will also have problems with rendering API unless it'll be served by local server (browsers forbid loading XHR via `file://` protocol).
2019-09-11T175832

Did we decide it we remove downloads retroactively, or will they be disabled from version 10.0 going forward? We should check if we can configure something like that in RTD service or propagate these changes from oldest version and rebuild them all.

RomanSek:
Looks like we can only disable PDF and ePub. HTMLZip will always be available.
https://readthedocs.org/dashboard/piwik-pro-pdfd/advanced/

2019-09-11T180907

JacekClearcode:
> We'll have to decide what do we do with downloads section (https://readthedocs.org/projects/piwik-pro-pdfd/downloads/). I know that we planned to remove PDF and ePub, but HTMLZip will also have problems with rendering API unless it'll be served by local server (browsers forbid loading XHR via `file://` protocol).
> 2019-09-11T175832
> 
> Did we decide it we remove downloads retroactively, or will they be disabled from version 10.0 going forward? We should check if we can configure something like that in RTD service or propagate these changes from oldest version and rebuild them all.

I was thinking about hiding section from 10.0 onwards. But I think we can safely disable pdf and other things as well in older versions

JacekClearcode:
> It may be an issue with docker build. When I build it locally (without docker) it worked a lot better. 👍
> 2019-09-11T175127

Yeah I think this is a docker issue also I saw that there is a serious caching problem with api json files. Could you check them in the free time? @RomanSek

PR #249, opened 2019-09-30
PROD-1406 Fix documentation links

jzyinq:
Context:
https://clearcodehq.atlassian.net/browse/PROD-1406

It's a hotfix for our current version of documentation. 

Complete fix for this issue will be a part of another task:
https://clearcodehq.atlassian.net/browse/PPCDEV-10657

### DIFF
--- a/platform/authorized_api_guide.md
+++ b/platform/authorized_api_guide.md
@@ -299,9 +299,9 @@ There is no response example. API will return `204 No Content` status code.
 [Postman](https://www.getpostman.com/) is a multiplatform GUI application for creating API calls.
 PPMS allows you to export swagger documentation and easily import it to Postman.
 Depending of what you want to work with, you can import given swagger docs:
-* [Apps](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/9.0/platform/authorized_api/apps/public_v2.yaml)
-* [Users](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/9.0/platform/authorized_api/users/index.json)
-* [Access control](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/9.0/platform/authorized_api/access_control/public_v2.yaml)
+* [Apps](/_static/api/platform_apps_authorized_api.json)
+* [Users](/_static/api/platform_users_authorized_api.json)
+* [Access control](/_static/api/platform_access_control_authorized_api.json)
 
 Simply click in Postman: `import -> Import From Link`. Then all of your paths are imported!
 You have to override two things:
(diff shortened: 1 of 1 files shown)


platform/authorized_api_guide.md
```diff
@@ -299,9 +299,9 @@ There is no response example. API will return `204 No Content` status code.
 [Postman](https://www.getpostman.com/) is a multiplatform GUI application for creating API calls.
 PPMS allows you to export swagger documentation and easily import it to Postman.
 Depending of what you want to work with, you can import given swagger docs:
-* [Apps](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/master/platform/authorized_api/apps/index.json)
-* [Users](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/master/platform/authorized_api/users/index.json)
-* [Access control](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/master/platform/authorized_api/access_control/index.json)
+* [Apps](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/9.0/platform/authorized_api/apps/public_v2.yaml)
+* [Users](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/9.0/platform/authorized_api/users/index.json)
+* [Access control](https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/9.0/platform/authorized_api/access_control/public_v2.yaml)
```
RomanSek: Keeping track of it in next versions will be PITA. I think it would be better to link to static files in documentation.
Can you use same link as the one provided by OpenAPI?
2019-09-30T115216
If you need YAML (instead of JSON) you can place YAML in `_static` directory and link to it directly.

jzyinq: I don't need yaml here so changing urls here it's good enough. Great tip!

jzyinq: I've created similar PR to `10.0` to fix it for future releases https://github.com/PiwikPRO/PPMS-PublicDocs/pull/252.

RomanSek: Clicking these links moves user outside docs (he has to click back to return to docs). We used already [download](http://www.sphinx-doc.org/en/1.6/markup/inline.html#role-download) role in similar cases. See if it will match your use case.

RomanSek: I meant something like this:
```
:download:`referrer_type.json </_static/json/enum/referrer_type.json>`
```

jzyinq: That file is a `.rst` file, I don't see an option to use `:download:` in markdown. Could you give me an example of such a snippet? That supposed to be quick PR so maybe we could speed up a bit ;-)

RomanSek: I'm confused. I gave you link to docs and use case from out documentation. I also stated that we already use it in similar cases. You can see it here (line 45, near "Channel" text)
https://raw.githubusercontent.com/PiwikPRO/PPMS-PublicDocs/master/custom_reports/integrations/google_search_console.rst
If you don't want to use dowlnoad scheme it's fine, but I would like to get clear message what you want to achieve.

PR #259, opened 2019-10-09
PPROF-1531 add audience count history to protected AM api

kedod:
- Update public docs to support audience history count


kedod:
Closed due to missing few parameters. Will be reopened.

PR #362, opened 2020-03-05
PPTECHDOC-66 Added warning about Piwik Reporting API deprecation

zaan:


analytics/reporting_api.rst
```diff
@@ -1,5 +1,8 @@
-Reporting API
-=============
+Reporting API (deprecated)
+==========================
+
+.. warning:: On September '20 we’ll disable Analytics (classic) API for all cloud customers. The tracking API will remain as is. We recommend you to switch to Analytics (new) :ref:`custom-reports-http-api` as soon as possible. For more, see the `API migration guide <https://help.piwik.pro/support/migrate/analytics-new-api-migration-guide/>`_
```
RomanSek: Are we able to attach this change to the specific version number? Up to now we used following way of communicating such changes, but it's possible that this is special case.

https://github.com/PiwikPRO/PPMS-PublicDocs/blame/ada7ac046b8f331662379fab5bed9e4fd62e3358/analytics/tracker_snippet.rst#L39

RomanSek: Correction: We set there version that started to deprecate API, so in this case it would be v12.0. I can reformat that, but I'm not sure what is the plan for on-prem versions.

RomanSek: We can do that. So something like this?
```suggestion
.. deprecated:: 9.0.0
    This API will be disabled along with Analytics (classic). It will happen in version 13.0 (cloud) and next LTS version (on-prem). The tracking API will remain as is. We recommend you to switch to Analytics (new) :ref:`custom-reports-http-api` as soon as possible. For more info, see the `API migration guide <https://help.piwik.pro/support/migrate/analytics-new-api-migration-guide/>`_
```
It could help if we could specify LTS version.

zaan: I don't see the date when it will be changed. It's not tied to a given release, admins will be disabling an API gradually.
Next proposition:

On September '20 this API will be disabled on the cloud. Customers using PPAS on-premises will get it disabled with the next upgrade to 9.0.0 LTS or higher. 
The tracking API will remain as is. We recommend you to switch to Analytics (new) :ref:`custom-reports-http-api` as soon as possible. For more info, see the `API migration guide <https://help.piwik.pro/support/migrate/analytics-new-api-migration-guide/>`_

RomanSek: I created PR to v9.0 branch with this text (plus some formatting fixes).
As loong as cloud change won't be attached to any specific version this should be enough. I'm still worried that for LTS users this will be sudden change (without prior warning, but if they are informed about it by CS team this should be fine).
https://github.com/PiwikPRO/PPMS-PublicDocs/pull/363

RomanSek: If we want to add this info in v9.0, You can close this PR.

zaan: No worries, documentation is just an additional channel of communication but not a primary one ;) It won't be a sudden change, because on-prem customers have full control when they will upgrade.

zaan:
Closing pull request as changes were merged with another pull request to 9.0 version.

PR #363, opened 2020-03-10
PPTECHDOC-66 Added deprecation warning for reportning API

RomanSek:
2020-03-10T111510

### DIFF
--- a/analytics/http_api.rst
+++ b/analytics/http_api.rst
@@ -1,3 +1,5 @@
+.. _tracking-api-http:
+
 HTTP API
 ========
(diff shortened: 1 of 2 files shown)


RomanSek:
2020-03-11T114152
Let me know if this is the look we're going for.

zaan:
Looks good!
I think that it would be good to add some separator between deprecation notice and the content of the docs or at least a bit more space.

RomanSek:
I don't see option to format it this way. Looks like only alternative format will  look this way:
2020-03-11T145727

zaan:
OK then, let's go with the first option with spaces between paragraphs.

PR #365, opened 2020-03-11
10.0

RomanSek:


_static/css/custom_1582183857228.css
```diff
@@ -235,3 +233,20 @@ redoc .param-name {
 	word-break: break-all;
 	white-space: normal !important;
 }
+
+body {
+    overflow-x: initial !important;
```
RomanSek: @JacekClearcode It's standard upstream merge. Is there a reason why v10.1 doesn't use new redoc, or can I merge it safely?

custom_reports/columns.rst
```diff
@@ -96,315 +96,297 @@ The table below lists core dimensions that may be used in queries.
 
 .. table:: Base Dimensions
 
-    +--------------------------------------+-----------------------------------+----------+--------+--------------------------------------------------------------------------------------------------+
-    |            Dimension Name            |             Column ID             |   Type   |Nullable|                                            Enum Table                                            |
```
RomanSek: @nfpp It's standard upstream merge. Is there a reason why v10.1 doesn't use updated columns, or can I merge it safely to match those from v10.0?
v10.0: https://developers.piwik.pro/en/10.0/custom_reports/columns.html
v10.1: https://developers.piwik.pro/en/10.1/custom_reports/columns.html

nfpp: @RomanSek  I don't understand this diff, locally I copied this table to separate files from those two branches and do `diff 10.0.txt 10.1.txt` and there are no changes. Please check after the merge that it will stay this way.

_static/css/custom_1582183857228.css
JacekClearcode: As we spoke privately, you can merge this safely.

PR #386, opened 2020-04-02
WIP PPNA-2579 goals api

kedod:
Wait for 14.0


nfpp:
Could you please ad any link with generated docs? It would be easier to review.

PR #457, opened 2020-08-05
add app with all fields

chriskicker:


chriskicker:
> I'm only wondering why destination of this PR is v12.0. Is it fix for that and all following version?

yup, exactly

RomanSek:
Merge it when you'll be ready and I'll merge it upstream later

PR #499, opened 2020-10-02
TM-5782: Tag Manager API documentation

rtreffler-piwik:
- adds Tag Manager API v1 documentation
- improves convert.py to handle lists
- improves convert.py to handle path/to/file.json#definition/reference

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,6 @@ xml_output
 private_*
 .rope_project/
 readthedocs/htmlcov
-tags
 .python-version
 *.pyo
 .pytest_cache/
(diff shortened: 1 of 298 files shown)


JacekClearcode:
image
So it doesn't work

rtreffler-piwik:
> image
> So it doesn't work
> https://developers.piwik.pro/en/feature-tm-5782-tm-api-documentation/tag_manager/authorized_api/tags.html
> 
> File exists at the https://developers.piwik.pro/en/feature-tm-5782-tm-api-documentation/_static/api/tag_manager_authorized_api_variables.json and you expected it to be at
> https://developers.piwik.pro/en/_static/api/tag_manager_authorized_api_versions.json
> so this is problem with url that you set for the redoc to pull that json file.

thanks for checking, can you see it now ? new commit should fix that issue

JacekClearcode (approved):

bua89 (approved):

PR #500, opened 2020-10-05
TM-6612: Example in create tag public api documentation is not loading

rtreffler-piwik:
- improves convert.py to handle OpenAPI 3.0 external examples (externalValue)

### DIFF
--- a/convert.py
+++ b/convert.py
@@ -78,6 +78,10 @@ def resolve(content, dir):
             tmp = load_content(file_path)
             content.update(resolve(search(resolve_local(tmp[0]), ref), os.path.dirname(file_path)))
             del content['$ref']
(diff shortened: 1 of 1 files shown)


JacekClearcode:
To check if it works: https://developers.piwik.pro/en/bugfix-tm-6612-example-in-create-tag-public-api-documentation-is-not-loading/tag_manager/index.html

rtreffler-piwik:
> To check if it works: https://developers.piwik.pro/en/bugfix-tm-6612-example-in-create-tag-public-api-documentation-is-not-loading/tag_manager/index.html

works splendidly, thanks

JacekClearcode (approved):

PR #507, opened 2020-10-23
PPTT-1362 Added ping method

RomanSek:
DO NOT MERGE TO 15.1 (changes for next version)

Preview:
https://developers.piwik.pro/en/feature-pptt-1362-added-ping-method/tracker/js_tracking_api.html#send-ping-request-on-demand
https://developers.piwik.pro/en/feature-pptt-1362-added-ping-method/tracker/tracker_object.html#ping

### DIFF
--- a/tracker/tracker_object.rst
+++ b/tracker/tracker_object.rst
@@ -569,6 +569,10 @@ Anonymization
 Advanced Usage
 --------------
 
+.. function:: ping()
+
+    Ping method may be used to update session dimensions during page view without attaching them to actions.
+
 .. function:: addListener(domElement)
 
     The function will add a click listener to link element.
--- a/tracker/tracker_object.rst
+++ b/tracker/tracker_object.rst
@@ -571,7 +571,7 @@ Advanced Usage
 
 .. function:: ping()
 
-    Ping method may be used to update session dimensions during page view without attaching them to actions.
+    Ping method sends request that will update session values without creating new event or page view. Most common use for this method is update of session custom dimensions or custom variables.
 
 .. function:: addListener(domElement)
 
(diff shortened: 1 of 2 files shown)


RomanSek:
@JacekClearcode @wronan can you review this PR? W want to add this feature in 15.2.

JacekClearcode (approved):

tracker/tracker_object.rst
```diff
@@ -569,6 +569,10 @@ Anonymization
 Advanced Usage
 --------------
 
+.. function:: ping()
+
+    Ping method may be used to update session dimensions during page view without attaching them to actions.
```
wronan, line 574: Don't we want to list what dimensions are updated? What are the most typical use cases?

RomanSek: I think it may be described too briefly. It should be
```suggestion
    Ping method may be used to update session custom dimensions during page view without attaching them to actions.
```

wronan: That better but still not entirely true, because it updates not only custom dimensions.
I'm not saying my version is perfect but here's what I've commited lately: https://github.com/PiwikPRO/PPMS-PublicDocs/blob/15.2/tracker/index.yaml#L660-L663 - maybe it will inspire you. Or maybe just copy-paste it here. Do as you like.

RomanSek: OK, maybe this will be better description:
```suggestion
    Ping method sends request that will update session values without creating new event or page view. Most common use for this method is update of session custom dimensions or custom variables.
```
Other uses for HTTP API ping (like Page Performance Metrics, prolong the session) doesn't apply here, since we have dedicated methods to send such ping requests (like automatic gathering navigation timing page performance metrics at the initiation of the tracker and heart beat setup). This API is dedicated only to allow user to force tracker to send ping request and update of custom dimension values was genesis for this feature.

wronan: Fine by me.

wronan (approved):

PR #524, opened 2021-01-25
TM-7033: Document datalayer events needed to do CM analytics

musialp:
Provide documentation for data layer events for CM analytics

### DIFF
--- a/consent_manager/insights_analytics/data_layer_events.rst
+++ b/consent_manager/insights_analytics/data_layer_events.rst
@@ -1,0 +1,210 @@
+Consent Manager insights data layer events
+==========================================
+
+Introduction
+------------
+
+In the PPAS 15.3 release, a number of changes and additions have been introduced to the events and variables connected
+to Consent Manager's consent form behavior and visitors' interactions with the popup. These include:
+
+  1. Changes to the `Consent` variable
+  2. Deprecation of the `Consents were sent` event
+  3. New data layer events
+
+Changes to the `Consents` variable
+----------------------------------
+
+.. versionchanged:: 15.3
+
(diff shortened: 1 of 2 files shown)


consent_manager/insights_analytics/data_layer_events.rst
```diff
@@ -0,0 +1,210 @@
...
+
+  Event name
+
+Consent form interactions events
+````````````````````````````````
+
+For the purpose of tracking how visitors interact with the consent form, every button has an assigned data layer event,
+that is emitted each time it is clicked.
+
+Agree to all button
```
stiopanoff, line 145: Maybe we should add information about that event being emitted at clicking "Agree" at simple popup?

consent_manager/insights_analytics/data_layer_events.rst
```diff
@@ -0,0 +1,210 @@
...
+
+  {
+    event: 'stg.agreeToAllClicked',
+  }
+
+.. describe:: event
+
+  Event name
+
+Reject all button
```
stiopanoff, line 158: Maybe we should add information about that event being emitted at clicking "Disagree" at simple popup?

stiopanoff (approved):

consent_manager/insights_analytics/data_layer_events.rst
musialp: Good point, added.

consent_manager/insights_analytics/data_layer_events.rst
musialp: Good point, added.

stiopanoff (approved):

PR #534, opened 2021-03-05
TM-7303: Update Custom Consent Form example to use new tracking commands

musialp:
Update Custom Consent Form example to use new tracking commands


JacekClearcode:
For viewing pleasure: https://developers.piwik.pro/en/feature-tm-7303-consent-tracking-update-custom-form-example/

zawias-pro (approved):

PR #535, opened 2021-03-12
PPPSYS-21491 Fixed custom variable example in Tracker HTTP API

azarecki:


azarecki:
Screenshot - 2021-03-12T100212 828

prk3 (approved):

wronan (approved):

RomanSek (approved):
I'm a bit worried that someone may not read the important part, but alternative will look worse, so we can try this for now.
We also have docs for older versions of PPAS (from v6.0). This looks like something that could be updated through all these versions (or at least all currently used by our clients). What do you think?

azarecki:
We do not have customers on such versions so I think we do not have to do it for all. I would do it for the newest cloud and on-prem releases.

PR #536, opened 2021-03-12
PPPSYS-21491 Fixed custom variable example in Tracker HTTP API

azarecki:


wronan (approved):

sebastianpiskorski (approved):

tracker/index.yaml
```diff
@@ -103,11 +103,13 @@ paths:
               - value - Array with:
                 - (`string`) Custom variable name, max length: 200 characters
                 - (`string`) Custom variable value, max length: 200 characters
+
+            *IMPORTANT:* value should be urlencoded - example presented here is not encoded for readability reasons.
```
sebastianpiskorski, line 107: I would change `should be` to `have to be`. Not encoded input is invalid.

RomanSek:
I added it already using cherry-pick

PR #555, opened 2021-04-14
PPTT-1975 Added new custom dimension methods and addeddeprication war…

RomanSek:
Preview:
https://developers.piwik.pro/en/feature-pptt-1975-new-custom-dimention-methods/tracker/tracker_object.html#custom-dimensions
https://developers.piwik.pro/en/feature-pptt-1975-new-custom-dimention-methods/tracker/js_tracking_api.html#custom-dimensions

### DIFF
--- a/tracker/js_tracking_api.rst
+++ b/tracker/js_tracking_api.rst
@@ -434,6 +434,9 @@ Example of usage::
 
     _paq.push(["setCustomDimensionValue", 3, "loginStatus"]);
 
+Tracking Custom Dimension (legacy)
+``````````````````````````````````
+
 .. deprecated:: 15.3
     Function :func:`setCustomDimension` is deprecated due to difficulty of use (passed values should be URL encoded). Please use :func:`setCustomDimensionValue` instead.
 
@@ -472,6 +475,20 @@ You can access custom dimension by providing a function that will use the ``getC
     _paq.push([ function() {
         var customDimension = this.getCustomDimensionValue(index);
     }]);
+
+.. function:: getCustomDimensionValue(index)
+
+    :param number index: **Required** Index of custom dimension
+
+Example of usage::
+
+    _paq.push([ function() {
+        var customDimension = this.getCustomDimensionValue(1);
+        console.log(customDimension);
+    }]);
+
+Retrieving Custom Dimension (legacy)
+````````````````````````````````````
 
 .. deprecated:: 15.3
     Function :func:`getCustomDimension` is deprecated due to the difficulty of use (returned values should be URL encoded). Please use :func:`getCustomDimensionValue` instead.
--- a/tracker/js_tracking_api.rst
+++ b/tracker/js_tracking_api.rst
@@ -432,7 +432,7 @@ If you wish to set a custom dimension to use it in tracking functions, use the `
 
 Example of usage::
 
-    _paq.push(["setCustomDimension", 3, "loginStatus"]);
+    _paq.push(["setCustomDimensionValue", 3, "loginStatus"]);
 
 .. deprecated:: 15.3
     Function :func:`setCustomDimension` is deprecated due to difficulty of use (passed values should be URL encoded). Please use :func:`setCustomDimensionValue` instead.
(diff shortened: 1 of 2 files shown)


JacekClearcode (approved):

tracker/js_tracking_api.rst
```diff
@@ -410,6 +410,33 @@ Tracking Custom Dimension
...
+
+    When you set a Custom Dimension, that value will be used in all tracking requests within a page load.
+
+.. warning::
+    This function does not send any data to the :term:`Analytics`. It sets a Custom Dimension to be sent with following events (e.g. page view, ecommerce events, outlink or download events).
+
+
+Example of usage::
+
+    _paq.push(["setCustomDimension", 3, "loginStatus"]);
```
prk3, line 435: I think we should remove uses of deprecated methods from examples. This applies to the whole document.
```suggestion
    _paq.push(["setCustomDimensionValue", 3, "loginStatus"]);
```

prk3 (changes requested):

tracker/tracker_object.rst
```diff
@@ -206,24 +206,51 @@ Custom variables
...
 
     The function that sets a custom dimension to be used later.
 
     :param string customDimensionId: **Required** Id of custom dimension.
     :param string customDimensionValue: **Required** Value of custom dimension.
 
+.. function:: setCustomDimension(customDimensionId, customDimensionValue)
+
+    .. deprecated:: 15.3
+        Function :func:`setCustomDimension` is deprecated due to difficulty of use (passed values should be URL encoded). Please use :func:`setCustomDimensionValue` instead.
```
wronan, line 221:
```suggestion
        Function :func:`setCustomDimension` is deprecated due to the difficulty of use (passed values should be URL encoded). Please use :func:`setCustomDimensionValue` instead.
```

tracker/tracker_object.rst
```diff
@@ -206,24 +206,51 @@ Custom variables
...
+
+    The function that sets a custom dimension to be used later.
```
wronan, line 223:
```suggestion
    Function that sets a custom dimension to be used later.
```
Somehow I have a feeling that there should be no article at the beginning of this sentence (or maybe rather `a` instead of `the`), but I'm not sure. Consult someone if you feel like it.

tracker/tracker_object.rst
```diff
@@ -206,24 +206,51 @@ Custom variables
...
+    The function that will return the value of custom dimension.
+
+    :param string customDimensionId: **Required** Id of custom dimension.
+    :returns: Value set with :func:`setCustomDimensionValue`
+    :rtype: string
+
 .. function:: getCustomDimension(customDimensionId)
 
+    .. deprecated:: 15.3
+        Function :func:`getCustomDimension` is deprecated due to difficulty of use (returned values should be URL encoded). Please use :func:`getCustomDimensionValue` instead.
```
wronan, line 247:
```suggestion
        Function :func:`getCustomDimension` is deprecated due to the difficulty of use (returned values should be URL encoded). Please use :func:`getCustomDimensionValue` instead.
```

wronan (approved):

tracker/js_tracking_api.rst
RomanSek: I updated example for `setCustomDimensionValue` function, but I'm not sure we can remove other examples. The "JavaScript tracking API" page is mostly covered by description of command and examples of use since these are not standard JS functions and other example of use describes deprecated `setCustomDimension` command.

prk3: So you're saying that this page is not about `_paq.push(["setCustomDimensionValue"...` but rather javascript methods and that examples with deprecated `setCustomDimension` are fine because they are used to show analogies. Is that correct?

In that case I see two options:
1.Keep using `setCustomDimension` and only that one for consistency but mention that `setCustomDimensionValue` should generally be used instead.
2. Change all examples to `setCustomDimensionValue`

Either is fine for me, what matters is that people don't get confused because we're mixing both methods.

prk3 (approved):

PR #575, opened 2021-04-27
TM-7630: Upgrade API documentation (refs TM-7639)

bua89:
- provide template_options description
- mark spent time as deprecated spent_time
- provide debounced_history trigger type


bua89:
We can mark it as 15.3.3 version

musialp:
> We can mark it as 15.3.3 version

Patch versions are not marked in the documentation. Changing target branch to `15.3`.

musialp (approved):

tag_manager/authorized_api/src/tags/schema/tag_templates/piwik_attributes.json
```diff
@@ -107,9 +107,10 @@
                     "type": "boolean"
                 },
                 "spent_time": {
-                    "description": "Time spent on page threshold in seconds",
+                    "description": "Time spent on page threshold in seconds (is ignored and will be removed in next version)",
```
RomanSek, line 110: Does providing it in next version will generate error? Usually we give users more time to comply with the changes, but I don't think we have any standard for that.

tag_manager/authorized_api/src/tags/schema/tag_templates/ab_tasty_attributes.json
```diff
@@ -11,7 +11,7 @@
             ]
         },
         "template_options": {
-            "description": "Tag template options",
+            "description": "[Tag template options](#tag-template-options)",
```
RomanSek, line 14: Can you check on preview that these links work as intended? I'm not sure they do.
2021-04-27T154702

RomanSek (commented):
Preview:
https://developers.piwik.pro/en/feature-tm-7630-upgrade-api-documentation/tag_manager/authorized_api/tags.html

tag_manager/authorized_api/src/tags/schema/tag_templates/piwik_attributes.json
musialp: It will show up in the tag, so nothing will break on our side. As I understood the changes, it will be ignored by tracker. Am I correct? Next version here means API v2, which is not even planned yet.

tag_manager/authorized_api/src/tags/schema/tag_templates/ab_tasty_attributes.json
musialp: The preview works ok on my end. Any specific tag is selected (maybe dome templates were missed)?

bua89: There is an issue with empty template options. I am fixing this now.

bua89: Fixed

tag_manager/authorized_api/src/tags/schema/tag_templates/piwik_attributes.json
RomanSek: Currently we communicate to users that API version is same as PPAS version, this may cause confusion. Is there a way to explain it better?
2021-04-27T175238

RomanSek (approved):

PR #598, opened 2021-05-19
Pptt 2156 tracker raw data endpoint

wronan:


tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+          type: string
+          example: piwikpro.tld
+        scheme:
+          description: 'Scheme used for the request'
+          type: string
+          example: http
+        method:
+          description: 'Method used for the request'
+          type: string
+          example: GET
```
prk3, line 1415: This could be an enum.

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+            - - 'x-forwarded-for'
+              - '172.18.0.15'
+            - - 'accept-language'
+              - 'en-US,en;q=0.9'
+            - - 'user-agent'
+              - 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.3359.117 Safari/537.36'
+        query_params:
+          description: 'Tracking request parameters'
+          type: object
+          $ref: '#/components/schemas/TrackingParameters'
```
prk3, line 1433: Will query_params be deserialized or still in a string?

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+          example: 'particular-user_login'
+        body:
+          description: 'Request body payload'
+          type: string
+          example: "idsite=1&rec=0&action_name=This%20won%27t%20be%20recorded"
+        debug:
+          description: 'States whether request is used for debugging'
+          type: boolean
+          default: false
+          example: true
```
prk3, line 1450: I think booleans don't need examples.

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+    TrackingParameters:
+      description: A tracking log
```
prk3, line 1452:
```suggestion
      description: Parameters of a tracking request
```

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+      required:
+        - idsite
+      type: object
+      properties:
+        idsite:
+          description: ID of the tracked app/website
```
prk3, line 1458: Please add information that this can be uuid or legacy int.

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+          type: string
+          example: 4
```
prk3, line 1460: By looking at the example, one could think that it's a number. Let's quote numeric values if type is string.
```suggestion
          example: "4"
```

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+        url:
+          description: 'URL of the action'
+          type: string
+          example: 'https://piwik.pro/'
+        urlref:
+          description: 'HTTP referrer'
+          type: string
+          example: 'https://clearcode.cc/'
+        res:
+          description: 'Resolution of the visitor''s device in pixels'
```
prk3, line 1488: Use double quotes when string has single quotes.
```suggestion
          description: "Resolution of the visitor's device in pixels"
```

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+          type: string
+          example: '1920x1080'
+        idgoal:
+          description: 'Goal ID. Signifies that a goal has been reached and tracks its conversion.'
```
prk3, line 1492: Can be also uuid, right?

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+        _cvar:
+          description: '[Custom variables](https://help.piwik.pro/analytics/custom-variables/) set on the ``visit`` scope.'
+          type: string
+          example: '{"1":["AspectRatio","16:9"],"2":["Height","1080"],"3":["Width","1920"]}'
+        cvar:
+          description: '[Custom variables](https://help.piwik.pro/analytics/custom-variables/) set on the ``page`` scope.'
+          type: string
+          example: '{"1":["AspectRatio","16:9"],"2":["Height","1080"],"3":["Width","1920"]}'
+        bots:
+          description: 'Bots tracking enabled / disabled'
```
prk3, line 1520: This one is no longer used.

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+          type: string
+          example: 'searching for this'
+        search_cat:
+          description: 'Internal search category'
+          type: string
+          example: 'Result category'
+        search_cats:
+          description: 'Result categories'
+          type: string
+          example: 'res cat 1, res cat 2'
```
prk3, line 1679: I think this params should hold json string array.
```suggestion
          example: '["res cat 1","res cat 2"]'
```

tracker/tracker_debugger_api.yaml
```diff
@@ -1318,3 +1374,912 @@ components:
...
+          example: 'custom_dimension_3_value'
+        dimension4:
+          description: Custom dimension 4
+          type: string
+          example: 'custom_dimension_4_value'
+        dimension5:
+          description: Custom dimension 5
+          type: string
+          example: 'custom_dimension_5_value'
+        dimension6:
```
prk3, line 1906: We should stop at 3 or 5. 100 is unnecessary noise.

prk3 (approved):
Some small corrections needed.

tracker/tracker_debugger_api.yaml
wronan: As documentation states: object (meaning: deserialized to JSON).

tracker/tracker_debugger_api.yaml
wronan: You say "no examples for booleans". Someone else will then as "all parameters have examples - so why not those?"

Let's leave it as is. All params should have examples. It's important in a way, because that way I'm sure that I should expect `true` / `false` values rather then `1` / `0` or anything alike.

tracker/tracker_debugger_api.yaml
wronan: Yes it could... but for now it's just a string...

tracker/tracker_debugger_api.yaml
wronan: Well... it's of type "string" - we can type anything we want there. Not only int or a UUID. But since you ask - there you go.

tracker/tracker_debugger_api.yaml
wronan: Why?

wronan: I've escaped it the way documentation suggests. Even Swagger editor prefers single quotes.

tracker/tracker_debugger_api.yaml
wronan: Yes but after all it's a type "string" - it can be whatever sender's imagination produced.

tracker/tracker_debugger_api.yaml
wronan: Liar, liar...
Both [CF](https://github.com/PiwikPRO/PPMS-Trucker-CollectorFront/blob/master/src/tracker/query_params.rs#L42) and [CSA](https://github.com/PiwikPRO/PPMS-Trucker-CollectorStorageApi/blob/master/src/log_repository/logs.rs#L232) still support it, so it's possible it can pop up in the endpoint output thus should be documented.

tracker/tracker_debugger_api.yaml
wronan: I didn't find any "and so on up till X" option in OpenApi - and since we either document every possible parameter or we don't I decided to document those as well.

wronan: It produces noise - no doubt about it.
But removing them seems unfair.

RomanSek (approved):
Looks good to me.
Preview: https://developers.piwik.pro/en/pptt-2156_tracker_raw_data_endpoint/tracker/tracker_debugger_api.html
Merge if it looks good to you.

PR #602, opened 2021-06-11
PPTT-2207 more ping types

prk3:


tracker/tracker_debugger_api.yaml
```diff
@@ -225,9 +225,10 @@ components:
             - 16
             - 17
             - 18
+            - 19
+            - 20
+            - 21
```
RomanSek, line 230: I'm not sure if it's better to use enum here or int with min/max values
event_type

RomanSek (approved):

tracker/tracker_debugger_api.yaml
prk3: We'll have holes in the range, e.g. events 10 and 11. Thanks for reminding me though, I've commented these two out and added comments naming all event types.

PR #605, opened 2021-06-29
PPTT-2316 Added `page_url` and `event_type` parameters for `public_processor_storage_api_v1`

sebastianpiskorski:
Preview: https://developers.piwik.pro/en/pptt-2316-additional-debug-endpoint-parameters/tracker/tracker_debugger_api.html

### DIFF
--- a/tracker/tracker_debugger_api.yaml
+++ b/tracker/tracker_debugger_api.yaml
@@ -61,6 +61,61 @@ paths:
             example: john@doe.com
           in: query
           required: false
+        - name: page_url
+          description: Filters out sessions that don't contain at least one event with URL matching the specified string.
+          schema:
+            type: string
+            example: blog
+          in: query
+          required: false
+        - name: event_type
+          description: |
+            Filters out sessions that don't contain at least one event of the specified type.
+            The allowed values are:
+            - 1 # page view
+            - 2 # outlink
+            - 3 # download
+            - 4 # search
+            - 5 # custom event
+            - 6 # content impression
+            - 7 # content interaction
+            - 8 # goal conversion
+            - 9 # ecommerce conversion
+            - 12 # consent form impression
+            - 13 # consent form click
+            - 14 # consent decision
+            - 15 # custom ping
+            - 16 # cart update
+            - 17 # broken event
+            - 18 # excluded event
+            - 19 # heartbeat
+            - 20 # deanonymization
+            - 21 # page performance
+          schema:
+            type: integer
+            example: 1
+            enum:
+              - 1 # page view
+              - 2 # outlink
+              - 3 # download
+              - 4 # search
+              - 5 # custom event
+              - 6 # content impression
+              - 7 # content interaction
+              - 8 # goal conversion
+              - 9 # ecommerce conversion
+              - 12 # consent form impression
+              - 13 # consent form click
+              - 14 # consent decision
+              - 15 # custom ping
+              - 16 # cart update
+              - 17 # broken event
+              - 18 # excluded event
+              - 19 # heartbeat
+              - 20 # deanonymization
+              - 21 # page performance
+          in: query
+          required: false
         - name: lookup_window
           description: 'Describes how much back in time sessions shall be fetched, takes value in minutes'
           schema:


tracker/tracker_debugger_api.yaml
```diff
@@ -61,6 +61,41 @@ paths:
             example: john@doe.com
           in: query
           required: false
+        - name: page_url
+          description: Filter. A part of page URL of the event, that have to be in the session to display the session
```
prk3, line 65:
```suggestion
          description: Filters out sessions that don't contain at least one event with URL matching the specified string.
```

tracker/tracker_debugger_api.yaml
```diff
@@ -61,6 +61,41 @@ paths:
...
+          schema:
+            type: string
+            example: blog
+          in: query
+          required: false
+        - name: event_type
+          description: Filter. A type of the event, that have to be in the session to display the session
```
prk3, line 72:
```suggestion
          description: Filters out sessions that don't contain at least one event of the specified type.
```

tracker/tracker_debugger_api.yaml
```diff
@@ -61,6 +61,41 @@ paths:
...
+          schema:
+            type: int
```
prk3, line 74: Are you sure int is a valid type? https://swagger.io/docs/specification/data-models/data-types/#numbers
```suggestion
            type: integer
```

tracker/tracker_debugger_api.yaml
```diff
@@ -61,6 +61,41 @@ paths:
...
+            format: uint8
+            example: 1
+            enum:
+              - 1 - page view
```
prk3, line 78: Enum values should be integers too. The description is a comment for us.
```suggestion
              - 1 # - page view
```

prk3 (changes requested):
I'm sorry I'm so picky.

tracker/tracker_debugger_api.yaml
RomanSek: Yes, currently enum value is presented as a number with label pair. Enum should contain only numbers and explanation for them may be added in description if it's necessary.
2021-06-29T134934

RomanSek (changes requested):

tracker/tracker_debugger_api.yaml
sebastianpiskorski: It renders like this:
image

I'm not conviced that user should know only of possible values, since this is a public API. I think that if somebody would like to use this endpoint he should not look for the meaning of the values in other places or try to figure out those meanings from our UI.

sebastianpiskorski: Ok, i will try with the description.

wronan:
It's a public endpoint do I'd say describing it here is the right choice.
Obtaining the token is a separate problem and for sure should be described somewhere.

sebastianpiskorski:
@RomanSek I agree that this JWT description is pretty unclear. IMO since we do not decide how JWT works we should follow the platform way and maybe link to their guide: https://developers.piwik.pro/en/16.0/platform/authorized_api_guide.html , but this is a topic for another ticket.

BTW Should version 16.0, be already published?

prk3 (approved):

RomanSek:
> @RomanSek I agree that this JWT description is pretty unclear. IMO since we do not decide how JWT works we should follow the platform way and maybe link to their guide: https://developers.piwik.pro/en/16.0/platform/authorized_api_guide.html , but this is a topic for another ticket.
> 
> BTW Should version 16.0, be already published?

We can add that link if it's already described in docs.
Docs for version 16.0 is published, but it's still private, so there are no links to it yet, but if you enter correct URL in browser then you can see it.

tracker/tracker_debugger_api.yaml
```diff
@@ -61,6 +61,61 @@ paths:
...
+          schema:
+            type: string
+            example: blog
+          in: query
+          required: false
+        - name: event_type
+          description: |
+            Filters out sessions that don't contain at least one event of the specified type.
+            The allowed values are:
+            - 1 # page view
```
RomanSek, line 75: Looks better, but do we want to separate number from label with `#`?
2021-06-30T100434

sebastianpiskorski: it makes copy-paste easier

sebastianpiskorski: and about link to JWT, it is a subject for another ticket, this one is already overdue

sebastianpiskorski:
@RomanSek I'm not so sure about this hidden link:
image

JacekClearcode:
I disabled this warning for now... Well RTD done goofed this one.

RomanSek (approved):

PR #606, opened 2021-07-09
Feature/pptt 2314 urls

prk3:
Preview: https://developers.piwik.pro/en/feature-pptt-2314-urls/tracker/tracker_debugger_api.html

### DIFF
--- a/tracker/tracker_debugger_api.yaml
+++ b/tracker/tracker_debugger_api.yaml
@@ -234,6 +234,7 @@ components:
         - is_entry
         - event_type
         - event_url
+        - event_original
         - event_index
         - server_time
       type: object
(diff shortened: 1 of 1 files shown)


tracker/tracker_debugger_api.yaml
```diff
@@ -286,10 +287,15 @@ components:
           type: integer
           example: 8
         event_url:
+          format: url
+          description: A URL on which event occurred, human-readable
```
RomanSek, line 291: Looks good, but `human-readable` is very enigmatic way to distinguish transformation to UTF8 form od URI. We can leave it, but maybe we should define in glossary what it means and link there.

RomanSek (approved):

tracker/tracker_debugger_api.yaml
prk3: I've added an entry to our private glossary:
https://piwikpro.atlassian.net/wiki/spaces/PPTP/pages/85731116046/Glossary

PR #618, opened 2021-08-02
update authorized api doc

chriskicker:


platform/authorized_api_guide.md
```diff
@@ -1,13 +1,13 @@
...
+====================
 
 ## Introduction
 This page describes how to access Piwik PRO API which uses
 [client credentials](https://www.oauth.com/oauth2-servers/access-tokens/client-credentials/)
 OAuth grant type for obtaining user token.
 All data is sent and received as JSON and is compliant with [JSON API](http://jsonapi.org/) specification.
 
-### Obtaining token
+### Creating API Credentials
```
jzyinq, line 10: Looking at our current index I think we should have instead `Platofrm -> Authorized API guide` simple `Getting started` section that will be more intuitive to find if someone want to well... get starting using our API. 

When we get rid of Analytics classic this will be authorization solution for all documented endpoints so I don't see a reason we should hide that section in Platform module. 

image

jzyinq: BTW I found a golden link [regarding API Documentation & how to write such section](https://idratherbewriting.com/learnapidoc/docapis_doc_getting_started_section.html)

platform/authorized_api_guide.md
```diff
@@ -30,16 +30,18 @@ communication with API.
 
 Piwik PRO API tokens use [JWT](https://jwt.io/) format.
 
-Make POST call to `https://<domain>/auth/token` with header `Content-Type: application/json` and payload:
+Make POST call to `https://<domain>/auth/token` where  with header `Content-Type: application/json` and payload:
```
RomanSek, line 33: It doesn't look like a grammatical sentence to me.

RomanSek (approved):
I fixed shell example highlighting. Rest looks good, so you can merge it when you're ready.

PR #628, opened 2021-09-01
Bug/ppna 4014 fix api inconsistency

pulina:
https://developers.piwik.pro/en/bug-ppna-4014-fix-api-inconsistency/custom_reports/object_management_api/object_management_api.html

Old version for comparison:

https://developers.piwik.pro/en/16.0/custom_reports/object_management_api/object_management_api.html

### DIFF
--- a/custom_reports/object_management_api/definitions/custom_dimension.yaml
+++ b/custom_reports/object_management_api/definitions/custom_dimension.yaml
@@ -2,6 +2,7 @@ type: object
 required:
   - type
   - attributes
+  - id
 properties:
   type:
     type: string
@@ -19,7 +20,15 @@ properties:
       - name
       - active
       - case_sensitive
+      - scope
+      - website_id
+      - tracking_id
     properties:
+      website_id:
+        type: string
+        format: uuid
+        description: ID of the website.
+        example: dde2dc68-b6bb-45ee-9ac2-868b6181e33f
       name:
         type: string
         description: Name of the custom dimension.
@@ -30,9 +39,9 @@ properties:
       case_sensitive:
         type: boolean
         description: Specifies if custom dimension is case sensitive or not.
-      slot:
+      tracking_id:
         type: integer
-        description: Slot number for custom dimension.
+        description: Tracking id of custom dimension.
         minimum: 1
         readOnly: true
       extractions:
--- a/custom_reports/object_management_api/definitions/custom_dimension_create.yaml
+++ b/custom_reports/object_management_api/definitions/custom_dimension_create.yaml
@@ -15,7 +15,7 @@ allOf:
             enum:
               - session
               - event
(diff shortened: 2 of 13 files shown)


custom_reports/object_management_api/definitions/custom_dimension.yaml
```diff
@@ -19,7 +20,16 @@ properties:
       - name
       - active
       - case_sensitive
+      - scope
+      - website_id
+      - slot
```
MarcinMaslany, line 25: slot here, but is deleted below. 

It is read only so probably should be present, but definitely not required.

custom_reports/object_management_api/definitions/custom_dimension_create.yaml
```diff
@@ -7,16 +7,17 @@ allOf:
         type: object
         required:
           - scope
-          - website_id
+          - slot
```
MarcinMaslany, line 10: readonly field, so shouldn't be here probably

custom_reports/object_management_api/definitions/custom_dimension_details.yaml
```diff
@@ -7,7 +7,7 @@ allOf:
         type: object
         required:
           - scope
-          - website_id
+          - slot
```
MarcinMaslany, line 10: Does `required` in openapi have different meaning? It appears in every case here, so it doesn't look like a mistake

custom_reports/object_management_api/definitions/goal.yaml
```diff
@@ -1,15 +1,78 @@
...
-        required:
-          - website_id
-        properties:
-          website_id:
-            type: string
-            format: uuid
-            description: ID of the website.
-            example: dde2dc68-b6bb-45ee-9ac2-868b6181e33f
-  - $ref: '#/definitions/GoalDetails'
+      website_id:
```
MarcinMaslany, line 26: don't we want to use refs? We do have         `$ref: '#/components/schemas/WebsiteId'`

pulina: we have only `$ref: '#/parameters/WebsiteId'` but we could add components section in index.yaml.

custom_reports/object_management_api/definitions/custom_dimension_details.yaml
MarcinMaslany: yes, I was looking at wrong repo.

custom_reports/object_management_api/definitions/goal.yaml
MarcinMaslany: yes, I was looking at wrong repo.

MarcinMaslany (approved):

RomanSek (approved):

PR #659, opened 2021-09-27
PPTT-2753 Added a `User ID` guide

sebastianpiskorski:
[Preview](https://developers.piwik.pro/en/pptt-2753-user-id-guide/data_collection/web/guides.html)

### DIFF
--- a/data_collection/web/guides.rst
+++ b/data_collection/web/guides.rst
@@ -30,7 +30,7 @@ Page views
 Page view is the most basic type of the tracked event. It represents a single page viewing action.
 By default it's triggered only once as soon as the HTML content is loaded to the browser with the :ref:`trackPageView<jtc-api-trackPageView>` function.
 
-Example::
+.. code-block:: javascript
 
     _paq.push(["trackPageView"]);
 
@@ -40,8 +40,51 @@ Example::
 
 .. todo:: Explain how to change page URL and title for virtual page views or link to SPA instructions page.
 
+User ID
+-------
 
+`User ID` is an additional parameter, that allows you to aggregate data. When set you will be 
+able to search through sessions by this parameter, fiter reports through it or create Multi attribution reports
+using `User ID`. You can learn more about User ID `here <https://help.piwik.pro/support/getting-started/userid/>`_.
+To set up `User ID` with your JavaScript Tracking snippet follow this guide.
 
+To set up `User ID` parameter add a call to :ref:`setUserId<jtc-api-setUserId>` to your tracking code:
+
+.. code-block:: javascript
+
+    _paq.push(["setUserId", "user-name@example.com"]);
+
+.. note ::
+
+    #. Invoking ``setUserId`` won't send any tracking request. To add `User ID` to tracked data, you have to call ``setUserId`` before :ref:`trackPageView<jtc-api-trackPageView>`. It does not have to be the first one. Another way is to send ping request with :ref:`ping<jtc-api-ping>`, after setting the `User ID`.
+    #. `User ID` can't be longer than 1024 bytes. It will be 1024 characters if you use only ASCII characters, but Unicode characters may require more bytes per character, so you should make sure that Unicode identifier cut down to 1024 bytes is still unique.
+    #. `User ID` should be a uniquie value for each user. Otherwise metrics for different users might be merged in the reports.
+    #. Usually a `User ID` value is an user email, because this is the identifier that users use to log in to a website and it fulfils above requirements.
+
+It is a good practice to remove value of `User ID`, when the user logs out. Otherwise `User ID` value might affect session of other users, if they share the same device. To remove `User ID` value call :ref:`resetUserId<jtc-api-resetUserId>`.
+
+.. code-block:: javascript
+
+    _paq.push(["resetUserId"]);
+
+Full abstract example, might look like this:
+
+.. code-block:: javascript
+
+    var user = getUserData();
+    if (user.isLogged) {
+      _paq.push(["setUserId", user.login]);
+    } else {
+      _paq.push(["resetUserId"]);
+    }
+
+.. warning::
+
+    Do not unset `User ID` by setting it to some seemingly empty value, like ``_paq.push(["setUserId", " "]);`` or ``_paq.push(["setUserId", ""]);``. This way some value might be still send to Collecting & Processing Pipeline. What seems to be an empty value to a human, might be not true for a machine. Only using ``resetUserId`` will guarantee that no `User ID` value will be send.
+
+.. note::
+
+    Use of ``resetUserId`` is neccessary only when clicking on log out button does not result in a page reload. For example, when your page is a Single Page Application, or user logout is initiated by a widget and the widget does not cause the webpage to reload, then you have to call ``resetUserId``. Otherwise, when page reloads on logout, then a call to ``resetUserId`` is not a neccessity, but sill, a good practice.
 
 Custom Events
 -------------


data_collection/web/guides.rst
```diff
@@ -40,8 +40,47 @@ Example::
...
+
+To set up `User ID` dimension add a call to :ref:`setUserId<jtc-api-setUserId>` to your tracking code:
+
+.. code-block:: javascript
+
+    _paq.push(["setUserId", "user-name@example.com"]);
+
+.. note ::
+
+    #. `User ID` can't be longer than 1024 bytes, it will be 1024 characters unless you use only ASCII characters. When using Unicode check the length of user identifier, so you won't exceed 1024 bytes.
```
azarecki, line 56: I think this information should not be in a guide but in lower level -> function description.

conf.py
```diff
@@ -63,8 +63,7 @@
     'sphinx.ext.doctest',
     'sphinx.ext.intersphinx',
     'sphinx.ext.todo',
-    'sphinx.ext.mathjax',
-    'sphinx_tabs.tabs']
```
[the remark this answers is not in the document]
sebastianpiskorski: This `'sphinx_tabs.tabs'` crashes sphinx build

data_collection/web/guides.rst
sebastianpiskorski: It is requirement of our CPP, this is special dimension and it is better to warn user early than later looking up for issues "why I cant find users by their ID". It is an edge case, but our customers surprised us more than once.

data_collection/web/guides.rst
```diff
@@ -40,8 +40,47 @@ Example::
 
 .. todo:: Explain how to change page URL and title for virtual page views or link to SPA instructions page.
 
+User ID
+-------
+
+User ID is an additional session dimension in yours analytics data. When set you will be able to aggregate data by this dimension. You can use Tag Manger, default Piwik Pro tag to setup `User ID` without editing any JS code. To set it manually follow this guide.
```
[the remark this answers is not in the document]
sebastianpiskorski: I will add how other products user it, yet I thought that such description is for higher level articles.

data_collection/web/guides.rst
```diff
@@ -40,8 +40,47 @@ Example::
...
+
+It is a good practice to remove value of `User ID`, when the user logs out. Otherwise `User ID` value might affect session of other users. To remove `User ID` value call :ref:`resetUserId<jtc-api-resetUserId>`.
+
+.. code-block:: javascript
+
+    _paq.push(["resetUserId"]);
+
+Full abstract example, might look like this:
+
+.. code-block:: javascript
```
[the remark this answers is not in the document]
sebastianpiskorski: :blush:

conf.py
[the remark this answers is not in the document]
RomanSek: Try rebuilding docker container: `docker-compose up --build`

data_collection/web/guides.rst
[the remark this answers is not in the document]
RomanSek: If we have articles that explain it, we can only mention it here and link to them.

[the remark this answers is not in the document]
mgonera: I would refrain from using word `dimension` or `session dimension` as it may bring confusion with custom dimensions.

Another thing is about Tag Manager usage- this is hugely problematic cross-team issue. Overall as RevOps we don't advise to use User ID in Tag Manager, as customers are usually unable to configure it properly. They are not taking into account, that as effect of such a configuration, `setUserId` is invoked with each tracker execution. And when a situation comes that the User ID variable is not set (due to unavailability of element for example), user ID is cleared in such session. We already has such situations. My proposition for solving this was to change TM implementation but for now it's as it was.

data_collection/web/guides.rst
```diff
@@ -40,8 +40,47 @@ Example::
...
+
+    _paq.push(["setUserId", "user-name@example.com"]);
+
+.. note ::
+
+    #. `User ID` can't be longer than 1024 bytes, it will be 1024 characters unless you use only ASCII characters. When using Unicode check the length of user identifier, so you won't exceed 1024 bytes.
+    #. `User ID` should be a uniquie value for each user. Otherwise metrics for different users might be merged in the reports.
+    #. Usually a `User ID` value is an user email, because this is the identifier that users use to log in to a website and it fulfils above requirements.
+
+It is a good practice to remove value of `User ID`, when the user logs out. Otherwise `User ID` value might affect session of other users. To remove `User ID` value call :ref:`resetUserId<jtc-api-resetUserId>`.
```
mgonera, line 60: `value might affect session of other users.` + if they are sharing the device

data_collection/web/guides.rst
mgonera: I'd suggest to add important info: invoking `setUserId` won't send any info to the tracker on its own. It will be sent with next subsequent event. Do we want to include info that it can be submitted with ping?

mgonera (commented):
I left few suggestions

data_collection/web/guides.rst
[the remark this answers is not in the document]
mgonera: @RomanSek https://help.piwik.pro/support/getting-started/userid/

conf.py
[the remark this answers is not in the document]
sebastianpiskorski: restored

data_collection/web/guides.rst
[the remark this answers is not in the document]
sebastianpiskorski: - Changed dimension to parameter
- Added link to help center article
- Removed word `dimension`
- Removed reference to TagManager

data_collection/web/guides.rst
sebastianpiskorski: - Added info that you have send `User ID` with additional request
- Left info about `User ID` length

data_collection/web/guides.rst
sebastianpiskorski: added

data_collection/web/guides.rst
```diff
@@ -40,8 +40,51 @@ Example::
 
 .. todo:: Explain how to change page URL and title for virtual page views or link to SPA instructions page.
 
+User ID
+-------
+
+`User ID` is an additional parameter, that allows you to aggregate data. When set you will be 
+able to search through sessions by this parameter, fiter reports through it or create Multi attribution reports
+using `User ID`. You can :ref:`learn more about User ID here<https://help.piwik.pro/support/getting-started/userid/>` 
+To set up `USer ID` with your JavaScript Tracking snippet follow this guide.
```
RomanSek, line 49:
```suggestion
To set up `User ID` with your JavaScript Tracking snippet follow this guide.
```

data_collection/web/guides.rst
```diff
@@ -40,8 +40,51 @@ Example::
...
+    var user = getUserData();
+    if (user.isLogged) {
+      _paq.push(["setUserId", user.login]);
+    } else {
+      _paq.push(["resetUserId"]);
+    }
+
+.. warning::
+
+    Do not unset `User ID` byt setting it to some seemingly empty value, like ``_paq.push(["setUserId", " "]);`` or ``_paq.push(["setUserId", ""]);``. This way some value might be still send to Collecting & Processing Pipeline. What seems to be an empty value to a human, might be not true for a machine. Only using ``resetUserId`` will guarantee that no `User ID` value will be send.
```
RomanSek, line 83: Internally it is set to empty string ([ref](https://github.com/PiwikPRO/PPMS-Trucker-JS/blob/master/src/piwik.js#L4768)) and we [check](https://github.com/PiwikPRO/PPMS-Trucker-JS/blob/master/src/piwik.js#L3208) for empty value in code before using it.
```
this.resetUserId = function() {
    configUserId = '';
};
```

it's internal implementation but I expect that we'll always check for string values.
If we want to leave this warning, then maybe this will work better:

```suggestion
    Do not reset `User ID` by setting it to arbitrary empty value, like ``_paq.push(["setUserId", " "]);`` or ``_paq.push(["setUserId", ""]);``. While some of those examples may work in current version, only using ``resetUserId`` will guarantee that `User ID` value will be correctly removed in the future.
```

azarecki (approved):

data_collection/web/guides.rst
sebastianpiskorski: Not like this, we want users to use `resetUserId`, not to try to set empty value for themselves. We already had cases that user set some space or invisible unicode and told us that `User ID` do not work. Even if in reality `resetUserId` sets it t empty string, we want customers to use this very function to turn off `uid` parameter.

data_collection/web/guides.rst
sebastianpiskorski: fixd

data_collection/web/guides.rst
RomanSek:
```suggestion
    #. `User ID` can't be longer than 1024 bytes. It will be 1024 characters if you use only ASCII characters, but Unicode characters may require more bytes per character, so you should make sure that Unicode identifier cut down to 1024 bytes is still unique.
```

RomanSek (approved):

PR #684, opened 2021-10-19
PPCDEV-14236 removed user group modules endpoints

plks:
[Preview](https://developers.piwik.pro/en/bugfix-ppcdev-14236-modules-fixes/platform/authorized_api/modules/modules_api.html)


RomanSek (approved):
Still renders correctly. Please check if displayed endpoints are correct and you can merge it.

PR #706, opened 2021-10-29
PPTT-2911 removing non-existing parameter for tracker POST endpoint

wronan:


wronan:
Sorry guys. I know that the target should be specific version branch and knowing that I've checked out my working branch from `16.4`.
I admit that originally I've targeted `master` here by mistake but I was sure that right after that I've edited the PR. Apparently I didn't ...
Not it has correct target.

JacekClearcode (approved):
Now we can sleep at night. :trollface:

wronan:
Mistery solved: I did all that I've described here but for the other PR #705
Thanks for the late night action man! I owe you one...

PR #707, opened 2021-11-02
PPTT-2934 Added trackError method to the JTC

RomanSek:


data_collection/web/javascript_tracking_client/api.rst
```diff
@@ -2286,3 +2286,29 @@ Miscellaneous
...
                 jstc.enableJSErrorTracking();
+
+.. function:: trackError(message, url, lineNumber, column)
+
+    Attempts to send error tracking request using same format as native errors caught by :js:func:`enableJSErrorTracking`.
+    Such error request will still follow rules set for tracker, so it will be sent only when JS error tracking is enabled (:js:func:`enableJSErrorTracking` function was called before this attempt). It will also respect rules for tracking only unique errors.
+
+    :param string message: **Required** Error message
+    :param string filename: **Optional** Name of file in which error occurred
+    :param string|number lineNumber: **Optional** Line number of file in which error occurred
```
musialp, line 2297: Should `filename` and `lineNumber` be optional for this interface, if they are passed as the value of event action?

RomanSek: I didn't want to make usage of this feature more complicated then necessary (especially since getting error source is not that easy). In reports we'll still use same format, but missing parts of trace back will be represented as a `undefined` strings)

prk3 (commented):
We should also document an overload that takes one parameter of type Error and deduces line, column, filename from it.

RomanSek:
We'll add updated documentation for this feature when implementation will be accepted by TM team.

PR #730, opened 2021-11-16
Feature/pptt 2923 web log import update for new tracker

RomanSek:
[Preview](https://developers.piwik.pro/en/feature-pptt-2923-web-log-import-update-for-new-tracker/data_collection/other_integrations/web_log_analytics.html)


data_collection/other_integrations/web_log_analytics.rst
```diff
@@ -16,19 +16,23 @@ Sample command:
 
 .. code-block:: bash
 
-    import_logs.py --url=https://demo.piwik.pro --token-auth=XXXXXXXXXXXXXXXXXXXXXXXXXXXXX --enable-static --enable-bots --show-progress --idsite=X --recorders=2 --recorder-max-payload-size=50 sample.log
+    import_logs.py --url=https://demo.piwik.pro --client-id=*** --client-secret=*** --enable-static --enable-bots --show-progress --idsite=*** --recorders=2 --recorder-max-payload-size=50 sample.log
```
adaqus, line 19: Let's set `--recorder-max-payload-size` to 1 - otherwise `--enable-bulk-tracking` has to be used, which is not supported in new tracker.

adaqus (approved):

data_collection/other_integrations/web_log_analytics.rst
RomanSek: Actually `--recorder-max-payload-size` is ignored without `--enable-bulk-tracking` since GET requests don't have bulk option. We can remove that option from  example or add `--enable-bulk-tracking`.

adaqus: Let's just remove `--recorder-max-payload-size`.

PR #731, opened 2021-11-17
Update getting_started.md

Kwarks85:
Checking the process myself I was looking for "Profile" and then noticed it is (now) called "User panel"


RomanSek:
Please change the destination to first version that should have this change (e.g. 16.6).

RomanSek:
Thanks, for this PR, but we checked it in application and "User panel" is a name used in previous versions of PPAS (15.x). You can see documentation for that version [here](https://developers.piwik.pro/en/15.6/platform/authorized_api_guide.html).
We're working on providing PPAS version in UI to make it clear which version you are using in the future.

PR #740, opened 2021-11-23
Pptt 3047 fix parameter name

wronan:


RomanSek:
Is the destination version correct? I see that this parameter is available since 16.4.

wronan:
It is, but I thought that destination should always be the latest version. Shall I change to to `16.4` ?

JacekClearcode:
Yup you should, we can still access these older 16 + versions via url.

wronan:
Damn... I've checked out my working branch from the 16.6 branch. I'm closing this PR and will create another one because there is a chance that we will do one more rename.

PR #747, opened 2021-11-26
PPTT-3037 renamed rand parameter to r, updated PSA debugger docs

prk3:
Changes in debugger API have been copied from api docs:
https://piwikpro.atlassian.net/wiki/spaces/PPTP/pages/85949317734/Differences+in+visitors+recognition+between+regular+tracking+and+log+replay.?focusedCommentId=85949579323#comment-85949579323

### DIFF
--- a/data_collection/api/tracker_debugger_api.yaml
+++ b/data_collection/api/tracker_debugger_api.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.2
 info:
   title: Tracker API
-  version: 0.21.0
+  version: auto
   description: Exposes sessions
 paths:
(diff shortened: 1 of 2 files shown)


RomanSek (approved):
Looks good. Check [preview](https://developers.piwik.pro/en/bugfix-pptt-3037-r-param/data_collection/api/http_api.html) and merge.
One issue may be sudden change of parameter in API. I know that this is just cache buster, but it still should go through deprecation period for old parameter and we should migrate in that period use of that parameter in all clients.

sebastianpiskorski (approved):

PR #798, opened 2022-01-21
TM-9188 Update TM API docs

sksz:
- updated filters examples and divergence with JSON API

### DIFF
--- a/tag_manager/authorized_api/src/triggers.yaml
+++ b/tag_manager/authorized_api/src/triggers.yaml
@@ -1,10 +1,8 @@
 openapi: '3.0.0'
 info:
     title: Tag Manager API
+    description: In order to access those endpoints, you need to **authenticate**. <a href="../../platform/getting_started.html">Here is a guide on how to do it.</a>
     version: '1.0'
-
-security:
-    - BearerAuth: []
 
 servers:
     - url: https://example.piwik.pro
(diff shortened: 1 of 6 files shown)


bua89 (approved):

tag_manager/authorized_api/src/triggers.yaml
```diff
@@ -14,6 +14,16 @@ paths:
...
+            description: |
+                ### JSON API compatibility divergence
+
+                #### Filters
+                In JSON API standard (https://jsonapi.org/recommendations/#filtering) one can request elements with multiple
+                filter values by simply providing them separated by coma:
+
+                `triggers?filter[trigger_type]=click,event`
+
+                Unfortunately it is not possible in this endpoint.
```
RomanSek, line 26: I just want to make sure that this endpoint won't accept multiple values since other endpoints accept them with different format.

RomanSek (approved):
Check [preview](https://developers.piwik.pro/en/feature-tm-9188-update-ppms-apidocs-if-necessary/tag_manager/authorized_api/tags.html) and merge.

tag_manager/authorized_api/src/triggers.yaml
sksz: It won't accept multiple values. We've found a bug in this endpoint (https://piwikpro.atlassian.net/browse/TM-9231). It will be fixed as soon as possible :)

RomanSek: Then you get green light from me.

PR #856, opened 2022-03-29
[PPI-145] Add Piwik PRO SDK for Flutter

danieltwork:


JacekClearcode:
Only draft So we don't want it for 16.15? Or do you want to have it built for someone?

danieltwork:
> 

It is complicated. We want it to hit 16.15, but we can't do that until we publish the library. At the moment it is not possible to publish the library, we are waiting for verify pub.dev publisher.

JacekClearcode (approved):

PR #917, opened 2022-07-06
PPTT-3736 Documentation changes for log analytics

glothriel:


data_collection/other_integrations/web_log_analytics.rst
```diff
@@ -1,14 +1,18 @@
...
 .. default-domain:: bash
 .. _data-collection-web-log-analytics:
+.. _GitHub: https://github.com/PiwikPRO/log-analytics/
+
 
 Web Log Analytics
 =================
 
+
+Log analytics is a python script, that allows importing common web server log formats (nginx, apache, iss and more) directly to Piwik PRO. It's Free Software available under GPLv3 license, available on `GitHub`_
```
wronan, line 11:
```suggestion
Log analytics is a python script, that allows you to import logs of common web servers (nginx, apache, iss and more) directly to Piwik PRO. It's a free software available under GPLv3 license, available on `GitHub`_
```
Reason for the first suggestion: this compound noun was too complex - should be split.
Reason for the second suggestion: I'm not 100% sure but IMO the article should be there + I don't see the reason for the capital letters.

data_collection/other_integrations/web_log_analytics.rst
```diff
@@ -101,3 +105,29 @@ Supported log formats:
 - log files of some popular Cloud Saas services: Amazon CloudFront logs, Amazon S3 logs
 - streaming media server log files such as: Icecast
 - log files with and without the virtual host will be imported
+
+The script does not directly support importing logs from log aggregator tools, like Grafana Loki or ELK. If you'd like to import logs from one of those, you need to download them to the disk first.
```
wronan, line 109:
```suggestion
This script does not directly support importing logs from log aggregation tools, like Grafana Loki or ELK. If you'd like to import logs from one of those, you need to download them to the disk first.
```

data_collection/other_integrations/web_log_analytics.rst
```diff
@@ -101,3 +105,29 @@ Supported log formats:
...
+
+
+Performance considerations & rate limiting
+----------------------
+
+The script needs CPU to read and parse the log files, but it is usually Piwik PRO server itself which will limit the import speed due to network latency.
+To improve performance, you can use the **--recorders** option to specify the number of parallel threads which will import hits into Piwik PRO. We recommend to increase **--recorders=N** until import speed is satisfying for you.
```
wronan, line 116: I have sincere doubts about that question. Default settings allow to reach speeds over 1.5k req/sec (most of the time). I'd say that "we recommend not to touch the default values until you really need to" :)
If we really need to suggest any parameter value tweaks then my recommendation would be to keep the `--recorders` at max 2 and to play with `--recorder-max-payload-size` instead (untill they reach `max payload size` http error from the CollectorFront).

data_collection/other_integrations/web_log_analytics.rst
```diff
@@ -101,3 +105,29 @@ Supported log formats:
...
+
+If you are Piwik PRO Core user, please make sure, that you are not hitting rate limits, by using **--debug-force-one-hit-every-Ns** flag to slow down the import process.
```
wronan, line 118: I may be wrong here but the payload is not restricted by the rate limiter.
If so then keeping the `--recorders` low and `--recorder-max-payload-size` high will grant you a great throughput while keeping the actual requests to the tracker relatively low.
And this is the scenario we recommend.

wronan: Damn... I just realized that `--enable-bulk-tracking` is OFF by default (it was ON for Rancher version). @sebastianpiskorski  / @RomanSek do you remember why?

My general recommendation for any log import would be:
* --recorders=2
* --max-payload-size=1000
* --enable-bulk-tracking=1
Those are IMO the best values to start with for Promil setup.

data_collection/other_integrations/web_log_analytics.rst
```diff
@@ -101,3 +105,29 @@ Supported log formats:
...
+
+Your logs should be automatically rotated and stored on your webserver, for instance in daily logs
+*/var/log/apache/access-%Y-%m-%d.log* (where %Y, %m and %d represent the year,
+month and day).
+
+If so, you can import your logs automatically each day (at 0:30), by using a cron command:
+
+.. code-block:: cron
+
+        30 0 * * * /wherever-you-cloned-the-repo/log-analytics/import-logs.py -u example.piwik.pro `date --date=yesterday +/var/log/apache/access-\%Y-\%m-\%d.log`
```
wronan, line 133: `-u`? I'm not familiar with such option for the `log_import` script.
Besides, I won't oppose if you remove this whole section.

wronan (changes requested):
We should discuss what option values we suggest to the clients before we make it public. Let's wait for @sebastianpiskorski 's opinion on my last comments.

data_collection/other_integrations/web_log_analytics.rst
sebastianpiskorski: We disabled bulk tracking by default and by default it will run with `--recorders=1`, sow very very slow. So instead of recommending how to tweak recording speed I would just describe generally, how parameters that relates to recording speed works.

data_collection/other_integrations/web_log_analytics.rst
sebastianpiskorski: I think that we disabled `--enable-bulk-tracking`, because we mostly did so on Rancher, since it killed instances instantly. With new tracker, I think you have the point that we can provide default values tweaked already. So it would be:

--enable-bulk-tracking=1 - since it would be faster and new tracker should handle it eaisily
--recorders=4 - since on infras with HA this is CollectorFront max replicas value
--max-payload-size=95 - since this is max that can be send with each request maximized (POST limit payload is 6MiB and each request in bulk can have 64kiB, 6144÷64=96), minus 1 for additional JSON characters, like wrapping array, commas and so on.

sebastianpiskorski: As for limits on core it is 70req/s (it think that there no distinction between GET and POST), so `--debug-force-one-hit-every-Ns` would useful if we would change it to miliseconds, instead of seconds.

glothriel: To summarize, I changed the defaults in the script:
* we are using bulk tracking by default
* we are using max-payload-size=95
* we are using one recorder

Also I changed `--debug-force-one-hit-every-Ns` to `--sleep-between-requests-ms`

data_collection/other_integrations/web_log_analytics.rst
glothriel: removed

wronan (approved):

sebastianpiskorski (approved):

PR #952, opened 2022-10-24
Pptt 3801 goal UUID

wronan:


RomanSek (approved):

data_collection/api/tracking_api.yaml
```diff
@@ -343,11 +343,19 @@ paths:
...
             Goal ID. Signifies that a goal has been reached and tracks its conversion.
+            Accepts either Goal UUID or legacy integer Goal ID.
 
             Value ``0`` is reserved for E-commerce cart tracking.
           schema:
-            type: integer
-          example: 15
+            type: string
+            format: uuid
+          examples:
```
RomanSek, line 352: Example doesn't work in this format. Check it [here](https://developers.piwik.pro/en/pptt-3801_goal_uuid/data_collection/api/http_api.html).

See screenshot:

wronan: Yeah - I've noticed. But the syntax I've used is correct (with this OpenApi version at least). Swagged editor compiles it correctly. Can you "do something"  in the repo itself? :)

kororokke: I didn't saw any of this in html when I quickly glanced soooo prolly not. Besides we won't update react library if that's the case for now so I suggest other ways.

kororokke: Gonna additionally check it up tomorrow morning.

wronan: OK, I'll do some workaround (probably I'll put the legacy `int` version into the description).
I just want to mention that OpenApi 3.1 brought even more useful stuff (like multiple allowed types), it would be nice to utilize it someday 🤞🏻

wronan: But to be clear: the stuff I used is compliant with 3.0.x

kororokke: Yup, redoc update would be needed to show your examples there. Welp. Sooner than later.

wronan: Thanks for checking.

PR #964, opened 2022-11-22
TM-10063: Implement `lookup_table` variable type

archi-tektur:


kororokke:
https://developers.piwik.pro/en/feature-tm-10063-lookup-table/ - for checking if build properly

PR #973, opened 2023-01-26
Update content_security_policy.rst

rmastop-pp:
sync the codes as seen in the installation tab


musialp:
@rmastop-pp Since we have to make some adjustments to this document anyway, we are going to update the code in this PR https://github.com/PiwikPRO/PPMS-PublicDocs/pull/984. I am going to close this one, but thanks for the suggestion.

PR #980, opened 2023-02-20
PPTT-3900 JSTC tracking parameters for eCommerce 2.0

sebastianpiskorski:


kororokke:
Why `master` branch?

sebastianpiskorski:
@kororokke changed base branch to `feature/new-ecommerce`

data_collection/api/tracking_api.yaml
```diff
@@ -355,6 +355,18 @@ paths:
...
+          in: query
+          description: Type of event in tracking request. Used for some event types. Decides how other parameters will be interpreted.
+          schema:
+            type: integer
+            enum:
+              - 9 # Ecommerce conversion
+              - 16 # Cart update
+              - 22 # Product detail view
+              - 23 # Add to cart
+              - 24 # Remove from cart
```
RomanSek, line 368: Since it's a public API we should try to make the event types human readable to avoid confusion in the future.
Lets change the expected value here to a human readable name of the event type using kebeb-case.

PR #995, opened 2023-04-18
Pptt 4012 remove helpcenter duplicates

wronan:


kororokke:
Was it on separate page? If so.. i fear the 404 pages that will show when someone bookmarked this page

wronan:
Ania moved those docs to help.piwik.pro and asked me to remove the original once. I've offered to her some form of redirection (in case someone has bookmarked those docs on developers.piwik.pro) but she said that she want all trace of the removed from the developers.piwik.pro... so 404s are not our concerns.

kororokke:
:shrug: then These 404 looks like... *censored* but if you say so then... ;)

kororokke (approved):

PR #1009, opened 2023-05-28
Pppsys 40907 docs update

wronan:


kororokke:
17.3 ?

kororokke (approved):

RomanSek (approved):

RomanSek:
I think that it could go to the 17.2 since there we introduced ecommerce v2 and this instruction was incorrect there too.
We could also replace the instruction with the one from [help center](https://help.piwik.pro/support/getting-started/track-ecommerce/#ecommercecartupdate()):
> It should be called each time a returning visitor’s session starts, and only once per session. Its purpose is to restore the state of the cart that was abandoned in the previous session.

PR #1022, opened 2023-06-26
Drop "account name" from Next.js setup instructions

sjednac:
Adjust the configuration template to use `containerUrl` rather than `accountName`, which is marked as deprecated.

Use `container-id` and `container-url` as value placeholders (same as React and Angular examples).


danieltwork:
commit moved to  #1030

PR #1068, opened 2023-09-15
Ppcdp 1013

prk3:


kororokke (approved):

customer_data_platform/authorized_api/customer_data_platform_attributes_api.yaml
```diff
@@ -0,0 +1,298 @@
...
+            Aggregation type for attribute values.
+
+            Allowed values depend on selected `data_type`:
+
+            | data_type | first | last | first_last | list | list_unique | min | max | sum |
+            |-----------|-------|------|------------|------|-------------|-----|-----|-----|
+            | number    | ✓     | ✓    | ✓          | ✓    | ✓           | ✓   | ✓   | ✓   |
+            | bool      |       | ✓    |            |      |             |     |     |     |
+            | ip        |       |      | ✓          |      |             |     |     |     |
+            | datetime  |       |      |            |      |             | ✓   | ✓   |     |
```
azarecki, line 204: We should allow list of datetimes, I think list should be allowed for all. Is there some technical blocker or we just did not implement it? I think we should add, I will create task for it.

azarecki (approved):

customer_data_platform/authorized_api/customer_data_platform_attributes_api.yaml
azarecki: https://piwikpro.atlassian.net/browse/PPCDP-1463

prk3: There are no blockers. We only implemented these aggregations that were necessary for all predefined attributes.

PR #1091, opened 2023-11-07
PPNA-5915 exporter columns section

rafalmisiarek:
Exporter columns section:

image


kororokke (approved):

custom_reports/columns.rst
```diff
@@ -47,3 +47,62 @@ dimensions to metrics or different dimensions.
...
+| event_index                    | Starts from 0.                                        |
++--------------------------------+-------------------------------------------------------+
+| page_view_index                | Starts from 0.                                        |
++--------------------------------+-------------------------------------------------------+
+
+
+Sessions scope - preferred
+``````````````````````````
+
+.. include:: columns__exporter_preferred_sessions_columns.rst
```
mrfijas, line 88: would it be possible to include plain `json` file? it would be easier to generate and more universal. but if not, rst is ok :+1:

custom_reports/columns.rst
```diff
@@ -47,3 +47,62 @@ dimensions to metrics or different dimensions.
...
++--------------------------------+-------------------------------------------------------+
+| event_custom_dimension_1       | Number of custom dimension slots is configurable.     |
++--------------------------------+-------------------------------------------------------+
+| session_custom_variable_key_1  | Number of custom variable slots is configurable.      |
++--------------------------------+-------------------------------------------------------+
+| event_custom_variable_key_1    | Number of custom variable slots is configurable.      |
++--------------------------------+-------------------------------------------------------+
+| event_index                    | Starts from 0.                                        |
++--------------------------------+-------------------------------------------------------+
+| page_view_index                | Starts from 0.                                        |
```
mrfijas, line 81: I think these descriptions aren't for exporter, but for internal columns - there's:
1. `website_name__id/website_name__label` instead of `app_uuid`
2. `timestamp` instead of `server_time`
3. `ipv4/6_address` instead of `location_ipv4/6`

rest of them seems alright

conf.py
```diff
@@ -71,7 +71,8 @@
     'sphinx.ext.todo',
     'sphinx.ext.mathjax',
     'opengraphext',
-    'sphinx_tabs.tabs']
+    'sphinx_tabs.tabs',
+    'sphinx_copybutton']
```
mrfijas, line 75: how does it work? it adds copy button to all `code-block`s?

rafalmisiarek: yes, it's copy button in upper right corner

custom_reports/columns.rst
rafalmisiarek: It will change nothing as still `.. code-block:: json` will be needed

mrfijas: but cannot it be something like
```suggestion
.. code-block:: json
    "columns": .. include:: columns__exporter_preferred_sessions_columns.json
```
? most probably can't, but worth asking :smile:

custom_reports/columns.rst
```diff
@@ -47,3 +47,62 @@ dimensions to metrics or different dimensions.
...
+The tables in the following sections shows all columns that may be easily used in Exporter config.
+For sessions and events scope there is preferred list of columns due to performance and reliability reasons.
+
+.. note::
+    The table below presents a short description of the selected fields.
+
++--------------------------------+-------------------------------------------------------+
+| Column                         | Description                                           |
++================================+=======================================================+
+| website_name                   | Binary format (16 bytes).                             |
```
mrfijas, line 63: I'm afraid we misunderstood - these descriptions aren't for exporter.
in exporter `website_name` and locations aren't in binary format, timestamp is not in UTC.. tbh I'd remove this whole section as it doesn't make too much sense to me

pulina (approved):
Remove this section and we good to go

custom_reports/columns.rst
rafalmisiarek: Nope, `..include` needs to be first, generally it can't be nested this way

custom_reports/columns.rst
rafalmisiarek: removed

mrfijas (approved):
only comment in https://github.com/PiwikPRO/PPMS-Analytics-Reporting/pull/2670#discussion_r1386405393

PR #1094, opened 2023-11-24
Update public users API

plks:
Change of list of supported languages


kororokke (approved):

kororokke:
I would prefer to target next version that is around the corner, instead of exisiting right now.

plks:
we try to be strict as it's meant for 18.6

PR #1122, opened 2024-02-12
Minimal password length changed to 12

plks:
Update public_v2.yaml


jzyinq (approved):

jzyinq (commented):
@plks this is worth a mention in public changelog as well imo

https://piwik.pro/changelog/version-18/

PR #1128, opened 2024-02-27
PPNA-6416 Real-time in Columns (scope -> scopes, Real-time, Query API, Raw data API)

mrfijas:


pulina (approved):
I did not check every record to validate is scope are correct. But  brief verification of few picked up sames to be good forecast.

custom_reports/integrations/sharepoint.rst
```diff
@@ -3,6 +3,7 @@ SharePoint
 
 Once SharePoint integration is enabled, additional metrics and dimensions will
 become available in the :ref:`custom-reports-http-api`.
+Sharepoint data isn't available in Real-time API.
```
pulina, line 6: Is this true? Doeas this mean event stream will not have 11 event type  al tall or we will not have additional sharepoint dimensions there?

pulina (approved):

custom_reports/integrations/sharepoint.rst
mrfijas: there's no Sharepoint dimensions in Real-time and [Sharepoint is not sold](https://piwikpro.slack.com/archives/C1PH7HZFT/p1707813012073119?thread_ts=1707744657.966529&cid=C1PH7HZFT) anymore, so there should be no new events with sharepoint type

janusz-piechota-piwik (approved):

PR #1129, opened 2024-03-01
PPNA-5383 add units to session_total_time and timing_dom_interactive …

pulina:
https://piwikpro.atlassian.net/browse/PPNA-5383


pulina:
Should be done in docs generator

PR #1130, opened 2024-03-04
PPNA-6416 Real-time query API

mrfijas:
Copy/paste of query API with `date_from/date_to -> lookback_window` change

### DIFF
--- a/custom_reports/http_api/definitions/real_time_query_request.yaml
+++ b/custom_reports/http_api/definitions/real_time_query_request.yaml
@@ -1,0 +1,253 @@
+type: object
+properties:
+  website_id:
+    type: string
+    format: uuid
+    description: ID of a website or a meta site.
+    example: 70236925-724f-404e-b634-7b5633bc0bcc
+  columns:
+    type: array
+    description: Column definitions for the query.
+    items:
+      type: object
+      properties:
+        column_id:
+          type: string
+          description: ID of the column (either a dimension or a metric).
+        transformation_id:
+          type: string
+          description: |
+            Optional ID of the transformation. Must be supported by the
+            column that was specified.
+        goal_id:
+          type: integer
+          description: |
+            Optional ID of a goal. Available only for metrics related to goals.
+      required:
+        - column_id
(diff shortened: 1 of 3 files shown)


custom_reports/http_api/definitions/real_time_query_request.yaml
```diff
@@ -0,0 +1,253 @@
...
+            Optional ID of the transformation. Must be supported by the
+            column that was specified.
+        goal_id:
+          type: integer
+          description: |
+            Optional ID of a goal. Available only for metrics related to goals.
+      required:
+        - column_id
+    example: [{"column_id":"device_type"},{"column_id":"revenue","transformation_id":"sum"}]
+  lookback_window:
```
pulina, line 29: Now I have few seconds to wander why we have `lookback_window` here? We have it case we implement it this way here https://github.com/PiwikPRO/PPMS-Analytics-Reporting/commit/f12eb73fe3512bc0563cea8af599ccb170e289a4. But on grooming we [decided to go with date_time approach](https://piwikpro.atlassian.net/wiki/spaces/PPNA/pages/86346957059/2024-01-30+Public+API+in+RTR). Is this bug  that slipped in or we have some further discussion and make different decisions.

pulina (approved):

custom_reports/http_api/definitions/real_time_query_request.yaml
mrfijas: raw data API uses date_from/date_to, query API uses lookback_window

custom_reports/http_api/definitions/real_time_query_request.yaml
```diff
@@ -0,0 +1,253 @@
...
+    type: int
+    format: date
```
janusz-piechota-piwik, line 31: Is it correct ? Format date is for string (for example, 2017-07-21) https://swagger.io/docs/specification/data-models/data-types/#format

mrfijas: of course not, good catch :+1:

janusz-piechota-piwik (approved):

PR #1140, opened 2024-03-26
PPCDP-2053 Removed deprecated CDP Activations API

azarecki:
@kororokke could you please create preview?


azarecki:
@kororokke @jzyinq please review

jzyinq (approved):
lgtm :ok_hand: but did we inform clients earlier that those endpoints will be unavailable from this release? Or do we just hide them from public documentation in this step?

azarecki:
It the documentation we informed that the API will be removed in 18.14. We also checked if anyone uses the API and Grafanas are empty. We are removing the API in this release

PR #1141, opened 2024-04-02
PPNA-6403 Update documentation based on date_from/date_to API specifi…

janusz-piechota-piwik:
…cation
obraz
obraz
obraz
obraz
obraz
:/

### DIFF
--- a/custom_reports/http_api/definitions/query_request.yaml
+++ b/custom_reports/http_api/definitions/query_request.yaml
@@ -28,14 +28,22 @@ properties:
     example: [{"column_id":"device_type"},{"column_id":"revenue","transformation_id":"sum"}]
   date_from:
     type: string
-    format: date
-    description: Start date for the query (inclusive). Cannot be used with relative_date field at
-      the same time. Mandatory if ``relative_date`` is not used.
+    oneOf:
+      - format: date
+        description: Start date for the query (inclusive).
+      - format: datetime
+        description: Start datetime for the query (inclusive, in website's timezone).
+    description: Cannot be used with relative_date field at the same time.
+      Mandatory if ``relative_date`` is not used.
   date_to:
     type: string
-    format: date
-    description: End date for the query (inclusive). Cannot be used with ``relative_date`` field at
-      the same time. Mandatory if ``relative_date`` is not used.
+    oneOf:
+      - format: date
+        description: End date for the query (inclusive).
+      - format: datetime
+        description: End datetime for the query (exclusive, in website's timezone).
+    description: Cannot be used with ``relative_date`` field at the same time.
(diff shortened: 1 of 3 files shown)


custom_reports/http_api/definitions/query_request.yaml
```diff
@@ -28,14 +28,22 @@ properties:
     example: [{"column_id":"device_type"},{"column_id":"revenue","transformation_id":"sum"}]
   date_from:
     type: string
-    format: date
-    description: Start date for the query (inclusive). Cannot be used with relative_date field at
-      the same time. Mandatory if ``relative_date`` is not used.
+    oneOf:
+      - format: date
+        description: Start date for the query (inclusive).
```
mrfijas, line 33:
```suggestion
        description: Start date for the query (inclusive, in website's timezone).
```
dates are also in the timezone, and clients often ask about it

custom_reports/http_api/definitions/query_request.yaml
```diff
@@ -28,14 +28,22 @@ properties:
...
+      - format: datetime
+        description: Start datetime for the query (inclusive, in website's timezone).
```
mrfijas, line 35: if we use "datetime" here, it's worth unifying it in "Start date (with time) -> Start datetime" also. alternatively we could consult the copy with Joanna

kororokke: and also i would use ``` `relative_date` ``` where possible, in one place there is ``` `relative_date` ``` in both places, in some cases only in one place.

kororokke: Cannot be used with ``` ``relative_date`` ``` field at the same time.
      Mandatory if ``` ``relative_date`` ``` is not used.

vs

Cannot be used with relative_date field at the same time.
      Mandatory if ``` ``relative_date`` ``` is not used.

custom_reports/http_api/definitions/raw_data_real_time_request.yaml
```diff
@@ -28,11 +28,11 @@ properties:
   date_from:
     type: string
     format: datetime
-    description: Start date (with time) for the query (inclusive, in website's timezone).
+    description: Start datetime for the query (inclusive, in website's timezone).
   date_to:
     type: string
     format: datetime
-    description: End date (with time) for the query (exclusive, in website's timezone).
+    description: End datetime for the query (exclusive, in website's timezone).
```
janusz-piechota-piwik, line 35: @mrfijas is it now OK ?

mrfijas (approved):

pulina (approved):

kororokke (approved):

PR #1170, opened 2024-07-17
PPTT-4826 Device anonymization API changes

adaqus:


pmajocha (approved):

data_collection/api/tracking_api.yaml
```diff
@@ -791,6 +791,24 @@ paths:
 
             - ``1`` - all IP bytes will be masked (0.0.0.0), GeoIP data below Country level will be anonymized
             - ``0`` - available visitor data will be added to the session
+
+            You can additionally anonymize device dimensions/attributes and user agent by setting `dda` parameter to ``1``.
```
wronan, line 795: I wouldn't add such info for other parameters as this should rather be a strict documentation not a recipe. But if you think it indeed benefits others then leave it as is.

wronan (approved):

data_collection/api/tracking_api.yaml
adaqus: Good point. Removed.

PR #1182, opened 2024-09-30
middleware docs for 18.28

chriskicker:


platform/authorized_api/access_control/public_v2.yaml
```diff
@@ -1,4 +1,5 @@
 openapi: 3.0.3
+x-api-category: Access Control
```
jzyinq, line 2: That file update is not necessary since it only brings `x-api` tags that will be used with new Developer Portal, so it's safe to skip it entirely.

jzyinq (commented):
It seems that there are only formatting changes so it's not necessary to merge them at all imo.

platform/authorized_api/access_control/public_v2.yaml
chriskicker: and btw i see sth is wrong [here](https://docs.cldsrv.biz/master/promil/development/apis/http?spec=apps#operation/api_apps_details_v2), i see app details three times

chriskicker: and what about [that](https://github.com/PiwikPRO/PPMS-PublicDocs/pull/1182#discussion_r1782489895)?

jzyinq:
Closing as there are no notable changes :closed_book:

